PR #21, opened 2021-04-21
WCOR-000: Update drupal core to 9.1.7.

gatoona:
Update core to 9.1.7 and force version of core during build.


gatoona:
acquia/blt 12.8.0+ seems to be causing issues in pipeline. Requires composer 2, but pipelines only supports composer 1 at the moment. https://docs.acquia.com/resource/using-composer/

washco-or (approved):

PR #25, opened 2021-04-26
WCOR-0019799: Create resource content type from spec tool definition.

gatoona:
Delete plural resources type content. Create new singular resource type content. Enable Rabbit Hole and Pathauto module.


lgbridges (approved):
I saw that some additional modules being added were prerequisites for the story.  Not sure whether we want to keep addition of utility modules with global impact in a separate commit from the rest of the story, etc., or if that's being too picky.  Maybe Josh has some advice.

joshuami (approved):

joshuami:
I'm usually a fan of adding dependent modules as a part of feature stories rather than as separate stories. It makes the feature story a little bigger, but it also prevents having small stories that only have two files changed for enabling the module. A good rule of thumb for total number of files changed is whether the PR can still be easily reviewed. I consider this one pretty easy to review.

One note, the logic above would not apply to removing a module that made it all the way to production. To cleanly do that, you have to spit it into a PR to remove the dependent configuration and uninstallation separate from the PR to `composer remove` the module's code.

PR #33, opened 2021-05-04
WCOR-0019799: Complete resource content type according to spec sheet.…

gatoona:
… Default to frontpage view.


lgbridges (changes requested):
Looks pretty good, but couple suggestions. 

We'll want the content type general description to match what's in the spec tool: "Use <em>resources</em> to create a search result for websites and applications that are not content." The one that's in there now might be confusing for County staff editors.

Also, if you felt like changing your field display labels to sentence case ("Resource Description" >> "Resource description"), I wouldn't hate it. ;)

lgbridges (approved):

PR #34, opened 2021-05-05
WCOR-0019994: Shield module enablement.

gatoona:


gatoona:
**Changes made:**

1.  Add personal IDE configuration auto exports to .gitignore
2. Enable Shield module for all environments
3. Because credentials are stored in the config export for the module, workaround is to set it via EHV. Only environments that have this variable will prompt for a login.

lgbridges (approved):

PR #41, opened 2021-05-08
WCOR-0019897: Add alert content type.

jmilden-git:


lgbridges (changes requested):
There are some issues here that we need to address.  I'll follow up to schedule a time to talk.

lgbridges (changes requested):
Recreate Severity field as "List (integer)".

jmilden-git:
Ready for another attempt at PR approval.

lgbridges (approved):
Ready for QA.

Additional review and discussion with Josh regarding the numerous core.base_field_override files in the PR -- they are likely related to some enablement or disablement of translation "somewhere" (we couldn't figure out where / what caused it -- some global setting, or setting per content type, unclear) but Josh assured me these override definitions don't constitute a problem.  They would likely show up at some point anyway, and we'll be re-tweaking translation settings at some point to adjust things at a granular level as needed.

gatoona:
@lgbridges I'm sure it won't cause a problem to merge this and import the new config files, but I'm still hesitant to merge given that we're not 100% clear on what caused the change. Developer side of me is iffy on merging changes that we don't fully understand what happened. 

Given that this story is only for adding a simple alert type and we're not sure what happened to cause all these config files to be created, I'm not sure this should be merged.

gatoona (changes requested):
Working on cleaning up PR. Please hold on merging.

gatoona (dismissed):

washco-or:
Pipelines job failed due to a security requirement to update drupal/ctools. Update has been made in #46. Waiting for @lgbridges to merge that into develop.

gatoona:
Pipelines job failed due to a security requirement to update drupal/ctools. Update has been made in #46. Waiting for @lgbridges to merge that into develop.

jmilden-git:
> Pipelines job failed due to a security requirement to update drupal/ctools. Update has been made in #46. Waiting for @lgbridges to merge that into develop.

Can we cancel an old PR, this one that failed and just create a new one?

lgbridges:
> > Pipelines job failed due to a security requirement to update drupal/ctools. Update has been made in #46. Waiting for @lgbridges to merge that into develop.
> 
> Can we cancel an old PR, this one that failed and just create a new one?

Not sure if there's an advantage to doing that, and it could clutter our PR history further.  Plus, you separate the history of changes and discussion that has already happened on this PR page.

lgbridges (approved):

gatoona:
Looks like this failed due to that scheduler config file not being generated for the alert content type. 
This is because the content type was created before the scheduler module was enabled.

The fix would be to go the forms display for alerts content type, save, export that config file, and push back.

gatoona (approved):

lgbridges:
@jmilden-git -- 

QA passes according to test plan, but without some additional updates, I will have to create at least one additional story to fill in gaps.

Stale PR was refreshed from main Dev in GitHub; please be sure to 'git pull origin feature/STRY0019897' before continuing work.

Suggestions:

- Enable scheduler for this content type. Technically, this story passes without it since it's not written into the test plan, but it will have to be enabled for Alerts at some point. I think the issue was that Scheduler module hadn't made it to main Dev when you did this PR. It is active now, but not enabled for Alerts. If it doesn't get enabled with this story, a separate story will have to be created to do so, so you may want to enable it for this PR.

- I think Title is more of an administrative field for alerts and should probably not display in the alert banner. I would adjust the Alerts view to remove the Title field, leaving only 'Alert body' field.  That said, we could wait until themes show up to see what the adjustment should be.

- The view allows for one alert at a time, but we may want to allow more simultaneously (2 or 3?). Last September, we had COVID-19 and Wildfire alerts concurrently.

- The alerts are, as the test plan states, displaying for all "pages", or individual nodes. However, they are not appearing on view-pages (such as /news). Since I'm assuming many homepages or other key "pages" will actually be Views-based pages, this is something we may want to figure out now rather than creating a new story to address it.

- The block is currently set to display in "Highlighted" region, but I'm fairly certain that the incoming theme from I.F. will incorporate alerts and notifications into the Header region. Probably inconsequential until theme arrives, though.

PR #49, opened 2021-05-14
WCOR-0019898: Create and configure notifications.

lgbridges:


lgbridges:
I have commits from another PR here for some reason? Needs correction.

PR #52, opened 2021-05-19
WCOR-0019898: Create initial notification content type and block display

lgbridges:
**Use Case**

As a community member, I would like to see a web group specific notification banner in order to be informed of that web group(s) latest notices . As an editor with the appropriate role, I would like to be able to create web group specific notices and schedule start/end publication times.

**Implementation Details**

Create the notification content type and custom block view using the spec tool definition.

**Test Plan**

As an admin, I can create a notification, assign a group to it in "Display in" field, and it will display in a block in the header region for any page or news item. I can also set a publish and unpublish time in Scheduler Options for the notification, and when cron runs after the designated time, the publish/unpublish action runs. As an anonymous user, I can see any active notification on the header for certain pages or news items.

**Acceptance Criteria**

Test plan passes.

**Dependencies** 

Scheduler module - https://www.drupal.org/project/scheduler


gatoona:
Starting QA via ODE35

gatoona:
@lgbridges the only issue I ran into during QA is having the notifications show for its assigned group.
Problem seems to be from pathauto configurations.

Check out the new pattern set for the page content type. /admin/config/search/path/patterns
Instead of `[node:field_group]/news/[node:title]` it is now set to `[node:field_group:0:entity:field_group_path]/[node:title]`

This ensures the page uses only one and the short path from the group it references.

config/default/pathauto.pattern.news.yml
```diff
@@ -0,0 +1,22 @@
+uuid: fe87dbec-0ecd-4f8b-85d4-4152606eb12d
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+id: news
+label: News
+type: 'canonical_entities:node'
+pattern: '[node:field_group]/news/[node:title]'
```
gatoona, line 10: [node:field_group:0:entity:field_group_path]/news/[node:title]

config/default/pathauto.pattern.page.yml
```diff
@@ -0,0 +1,22 @@
+uuid: ec1ea3be-91cf-4cf8-ac2d-bac294eab081
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+id: page
+label: Page
+type: 'canonical_entities:node'
+pattern: '[node:field_group]/[node:title]'
```
gatoona, line 10: [node:field_group:0:entity:field_group_path]/[node:title]

config/default/pathauto.pattern.news.yml
gatoona, line 10: [node:field_group:0:entity:field_group_path]/news/[node:title]

gatoona (changes requested):

lgbridges:
@gatoona -- ODE35 updated with the suggested changes.  Please test at your convenience.

gatoona (approved):
@lgbridges Able to create notice, assign it to group(s), and have it show for content listed under that group(s).

PR #57, opened 2021-05-24
WCOR-0020262: Define the default rich text format

sharstine:


lgbridges:
Beginning code review and QA.

sharstine:
QA can begin on wcorode40.

gatoona:
@lgbridges looks like DB sync failed in ACP due to a difference in configuration in export and database.
Testing reveals it's a weird bug and not an issue with the code itself.

When filter.format.basic_html completes its initial create operation during configuration import, it is checked as different from the configuration in the database. Only when you import it again after the configuration has already been created in the database does the issue resolve itself.

So initial merge into develop for this PR will show an error in ACP, but will later be resolved during future PR merges.

Weird bug, but nothing we should worry about.

lgbridges:
@sharstine -- Code review and QA pass.  Please review the suggestions and let me know if this is good to merge or if you wish to make changes.

**Suggestion:**  The cut, copy, and paste toolbar buttons aren't necessary, but I know a surprising number of lower proficiency users who haven't memorized the keyboard shortcuts and rely on toolbar buttons for these functions.  We may want to add these.

**Suggestion:**  Since this is the initial enablement of image embeds in content, we may want to set a size restriction.  We're not concerned with site load time (we can employ a method like lazy loading as needed) so much as consumption of user bandwidth.  I have witnessed some authors embedding 10MB photos, so some restriction may be useful.  Open to discussing with the team.  My suggestion would be 1MB or 2MB.  2MB is a more generous limit, and very few people would have photos larger than this.  1MB may require working more closely with web groups on image optimization.

lgbridges (dismissed):
@sharstine -- Code review and QA pass.  Please review the suggestions and let me know if this is good to merge or if you wish to make changes.

**Suggestion:**  The cut, copy, and paste toolbar buttons aren't necessary, but I know a surprising number of lower proficiency users who haven't memorized the keyboard shortcuts and rely on toolbar buttons for these functions.  We may want to add these.  (Not referring to "Paste from Word", but just cut/copy/paste within RTE.)

**Suggestion:**  Since this is the initial enablement of image embeds in content, we may want to set a size restriction.  We're not concerned with site load time (we can employ a method like lazy loading as needed) so much as consumption of user bandwidth.  I have witnessed some authors embedding 10MB photos, so some restriction may be useful.  Open to discussing with the team.  My suggestion would be 1MB or 2MB.  2MB is a more generous limit, and very few people would have photos larger than this.  1MB may require working more closely with web groups on image optimization.

gatoona:
@sharstine @lgbridges for image uploads in ckeditor, should it not be a reference to a media image entity, and not a separate image uploader?

lgbridges:
> @sharstine @lgbridges for image uploads in ckeditor, should it not be a reference to a media image entity, and not a separate image uploader?

I think that's where we're headed, but having tested Media with this filter (switched on _Embed media_ option for the filter, configured it, added _Insert from Media Library_ toolbar button for CKeditor), the out-of-box library browser experience is seriously lacking. 

My previous experience with Media module is all coming back to me. The default "Add or select media" dialog, besides being a clunky library browser, doesn't seem to allow you to add an image on the spot (which is what 98% of authors do). This opens up a whole can of worms.

So first, we may have to [build a better media entity browser](https://www.ostraining.com/blog/drupal/select-embeddable-media/) that can embed media into the RTE. But even having reviewed this, I didn't see the "upload new image" issue addressed.

To me, this is the scope of a separate story. We have to research a solution for media library integration with RTE and media entity embeds, and implement it. I'm guessing that story could also address the file size limitation we've been discussing.

20210525a
20210525b
20210525c
20210525d
20210525e

sharstine:
@lgbridges @gatoona

**Regarding cut/copy/paste** - I agree these tools would be useful to include. I experimented with the standard "Paste" tool (which I envision being used the most often of those three tools), but it doesn't work in Edge, Firefox, or Chrome.

paste-message

While the "Paste" tool doesn't work, at least it tells the editor what they need to do. It's probably a good idea to add cut/copy/paste regardless. To add these tools, would I just do the following?

1. Make the changes on my local dev site
2. drush cex
3. git status
4. git add . (or replace . with the actual file updated)
5. git commit -m "WCOR-0020262: Enable cut, copy, and paste tools in CKEditor."
6. git push -u origin feature/STRY0020262

Anything I need to do after that?

**Regarding the media browser and image uploads** - I agree that a spike would be best. Sounds like some additional research is needed. The max file size limit adjustment could be coupled with the other media-related config changes, or I could just set it to 2 MB for now while adding the cut/copy/paste tools.

gatoona:
@lgbridges 

> The default "Add or select media" dialog, besides being a clunky library browser, doesn't seem to allow you to add an image on the spot 

I'm confused, the 'Add or select media' prompt does allow you to add an image on the spot. 
addnow

gatoona:
@sharstine 

> While the "Paste" tool doesn't work, at least it tells the editor what they need to do. It's probably a good idea to add cut/copy/paste regardless. To add these tools, would I just do the following?
> 
> Make the changes on my local dev site
> drush cex
> git status
> git add . (or replace . with the actual file updated)
> git commit -m "WCOR-0020262: Enable cut, copy, and paste tools in CKEditor."
> git push -u origin feature/STRY0020262

Yes, those order of commands would be correct.

sharstine:
Added cut, copy, and paste in CKEditor toolbar.

lgbridges:
@sharstine I heard that you will 

> 
> 
> @lgbridges
> 
> > The default "Add or select media" dialog, besides being a clunky library browser, doesn't seem to allow you to add an image on the spot
> 
> I'm confused, the 'Add or select media' prompt does allow you to add an image on the spot.
> addnow

Potential browser issue, then. I think I was testing in Edge.

lgbridges:
Beginning QA.

lgbridges (approved):

PR #61, opened 2021-05-26
WCOR-0020415: Install and enable S3Fs module.

gatoona:
### ODE42

Changes made:

- Install and enable S3FS module.
- Disable moderation for image media entities.
- Change storage class to S3 for image media entity.
- Enable display of error messages in Drupal UI.
- Include secrets.settings.php in settings.php.
- Update/New BLT commands to sync secrets.settings.php.

**Use Case:** As a content editor, I can upload images/documents via Drupal's UI and grab their URLs to use them for content.

**Implementation Plan:** Install and enable the S3FS Drupal module. Setup configuration and insert credentials. Link to temporary folder in existing S3 Bucket.

**Test Plan:** Upload a image file as a media entity. Check to see that the file has uploaded successfully and is viewable via its AWS URL.

**Acceptance Criteria:** Test plan passes.


sharstine:
Beginning QA.

lgbridges:
Reviewing code.

lgbridges (commented):
Code review passes.

sharstine (approved):
@gatoona @lgbridges

QA passes.

- Uploaded an image via /media/add/image.
- Inserted the uploaded image into a News node via the "Featured image" field.
- Image path points/redirects to an AWS URL.
- Uploading a new image version via /media/add/image updated the image src on the News node.

**Suggestion for a future story** - When you upload a new version of the image in Drupal, or delete the image in Drupal, the old version/file remains in S3 and is not deleted. As a follow-up to this story, perhaps we could look into a way to delete old image versions/files in S3 when they're deleted in Drupal.

lgbridges:
> **Suggestion for a future story** - When you upload a new version of the image in Drupal, or delete the image in Drupal, the old version/file remains in S3 and is not deleted. As a follow-up to this story, perhaps we could look into a way to delete old image versions/files in S3 when they're deleted in Drupal.

Created spike story STRY0020477 in backlog.

gatoona:
@lgbridges Images get deleted from S3 after 6 hours after being set to temporary in Drupal.

lgbridges:
> 
> 
> @lgbridges Images get deleted from S3 after 6 hours of being set to temporary in Drupal. Drupal deletes it this way to prevent accidental deletions if referenced by a content.

OK. 

> 
> 
> @lgbridges Images get deleted from S3 after 6 hours of being set to temporary in Drupal. Drupal deletes it this way to prevent accidental deletions if referenced by a content.

OK, that sounds good. Would it be worth our while to verify this now? @sharstine, any chance you recognize whether the items you deleted are no longer in the S3 bucket (see screenshot)?

Capture

gatoona:
> OK, that sounds good. Would it be worth our while to verify this now? @sharstine, any chance you recognize whether the items you deleted are no longer in the S3 bucket (see screenshot)?

@lgbridges I verified already in my local environment. Some of those items aren't going to get deleted because the environment Scott tested on got wiped out when I pushed my new commit. So Drupal has no record of those items after it got rebuilt pulling in the database from dev.

Here was my test in my local environment:

1. Uploaded a new image using media entity (drupal marks the file status as permanent)
2. Delete the media entity for the image (drupal marks the file status as temporary now)
3. After 6 hours, cron is ran and both mention of the file in Drupal and actual file in S3 is deleted.

PR #64, opened 2021-05-27
WCOR-0020478: Migrate departments into group content type.

gatoona:
### ODE45

**Changes made:**

- Enable CSV Importer.
- Define pattern for groups.
- Change field_type in groups to a select list according to spec sheet.

**Use Case**

As a authorized user, I would like to see departments listed under group types, under departments.

**Implementation Details**

Install and enable the csv_importer module. Import exported department csv data from central systems.

**Test Plan**

As an authorized user I can see all Washington County departments under content as a group, and am able to select a department for group fields. 

**Acceptance Criteria**

Test plan passes.

**Dependencies** 

CSV Importer module - https://www.drupal.org/project/csv_importer


lgbridges:
Starting QA and CR.

config/default/pathauto.pattern.group.yml
```diff
@@ -0,0 +1,22 @@
+uuid: 954cfb14-738d-4c26-afe2-2274e65cb3a5
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+id: group
+label: Group
+type: 'canonical_entities:node'
+pattern: '[node:field_parent:entity:field_group_path]/[node:field_group_path]'
```
lgbridges, line 10: I'm reading this pattern as /[group's parent group path name]/[group path name].  If that's right, have we determined that all groups with a parent should display that parent as part of the path?  We've talked a lot about creating a flatter hierarchy.  Would we want ITS to continue to appear as www/support-services/its?  Maybe so, but maybe not all divisions would be nested (like, I could envision Solid Waste just being www/waste rather than www/hhs/swr, etc)

Something for us to discuss further, won't hold this one up.

lgbridges (dismissed):
CR/QA pass.

**Further discussion:**  How will we structure group paths?  Setting the pattern to www/[parent group path]/[group path] is fine for a default setting initially.  We may go back and collapse some of those group paths individually if possible in cases where it makes sense to put that group's path right up under root.

config/default/pathauto.pattern.group.yml
lgbridges: Recognizing that this is just the default setting and we can adjust individual group URLs.

gatoona: @lgbridges I think we discussed having only a certain number of levels down in terms of the URL heiarchy. Noticing multco.us does 2 from the base (Ex. animalservices/licensing)

I haven't included it yet, but I was thinking we filter the field_parent to only display departments. That way we still keep a strict hierarchy system. Everything would basically be www/department/division, services, news, etc

washco-or: @gatoona test

lgbridges (approved):

PR #71, opened 2021-06-01
WCOR-0020526: Install and enable views data export module.

gatoona:
### ODE47

**Use case:** As an admin, I would like to be able to easily export Drupal data.

**Implementation:** Install and enable views data export module.

**Test plan:** As an admin, I can see that the views data export module has been installed and enabled. I am also able to export data in views using the module.


washco-or (changes requested):
Testing

lgbridges:
Starting QA.

lgbridges (approved):
Code review and QA pass.

PR #74, opened 2021-06-01
WCOR-0020523: Install and enable redirect module.

gatoona:
### ODE48

**Use case: ** As an admin, I would like to be able to set redirects on the site.

**Implementation:** Install and enable redirect module.

**Test plan:** As an admin, I can see that the redirect module has been installed and enabled. I am also able to set redirects on the site.


jmilden-git:
I'll attempt to do the QA for this one.

jmilden-git:
QA passed.

jmilden-git:
Looks like code review was done.  QA passed.  Showing a couple of conflicts.  Back to you @gatoona

lgbridges (approved):

PR #78, opened 2021-06-07
WCOR-0: Fix error when new content type is created.

gatoona:


lgbridges:
Patches an issue from [PR26](https://github.com/Washington-County/COUNTY-DRUPAL/pull/26) in which the core ```body``` field definition was wiped in favor of a custom ```field_body``` that would have no associated summary (we have a separate custom summary field). Core was issuing a whitescreen error for all new content types as a result (despite creating the content type). ```body``` as been added back to bypass the error, but developers should be aware that new content types now include the core ```body``` field -- it should be swapped for ```field_body``` on new content types.

lgbridges (approved):

PR #79, opened 2021-06-07
WCOR-0020518: Add document media type.

gatoona:
### ODE52

**Use case:** As an editor, I can upload a document and set metadata.

**Implementation:** Create document media type according to spec tool definition. Review and update file field storage pattern as necessary.

**Test plan:** As an authorized user, I can create a document, upload a file, and set associated fields.


jmilden-git:
I'll take QA.

jmilden-git:
@gatoona Ran into an error trying to upload a document.  Failed QA.  Not sure why.

gatoona:
@jmilden-git should be working now.

jmilden-git:
The document media type has passed QA.

jmilden-git:
@gatoona Not sure if someone else was doing code review.  I only saw one issue on a .yml file you might want to take a look at.

lgbridges (changes requested):
I didn't see any code _problems_ upon review, but some requested changes:

- Help text from spec tool is not included for _Show in media library_ and _Displayed in_ fields.
- Allowed file formats include TXT, DOC, DOCX, and PDF -- group is agreeing that PDF should be the only allowed format for Document media type.

@jmilden-git -- what was the code issue you referred to earlier, in one of the YML files?

lgbridges (approved):

PR #80, opened 2021-06-09
WCOR-0020675: Cleanup language selection fields and add in Spanish.

gatoona:
**Use case:** As an editor, I want to create translation pages for content.

**Implementation:** 

- Disable translation for interface items. No need to download and translate Drupal admin.
- Add Spanish as a translatable language.
- Enable translation detection through URL.
- Disable language selection field for content. (Defaults to English, then supports translations if added.)

**Test plan:** As an editor, I can create a page in English, and within the same node, provide non-English translations that can be accessed via the language code URL.

**QA:** ODE54


lgbridges (approved):
CR/QA pass.

Not specific to this story, but it's about time we open a new story to create some permissions upstream that enable content creation, editing, and translation, to make future testing easier and not have to toggle them on for each QA activity.

config/default/core.extension.yml
```diff
@@ -9,7 +9,6 @@ module:
   config: 0
   config_filter: 0
   config_split: 0
-  config_translation: 0
```
joshuami, line 12: @gatoona and @lgbridges , I wish ya'll had reached out to me on this one. You will need to both config_translation and locale even if you don't translate the admin side of the interface. Config translation and Interface Translation is what allows views and UI elements stored in config to get translation strings.

An interesting side effect of turning this off is that my config import for my story is failing now due to having changes that are incompatible with a site where those modules were turned off. (And here I thought I would get my story in under the wire 😄 ) Let's talk through this a bit in architecture tomorrow.

gatoona: @joshuami @lgbridges a few reasons why the config_translation module was disabled:

1. Interface translation can't be disabled because configuration translation requires it.  When adding additional languages with interface translation enabled, Drupal automatically downloads core translation kits relative to the language added which added considerable time to the build process. Just for Spanish alone, it took 3 mins to download and install its language kit.
2. We have successfully completed a Drupal site with translations before without configuration translation enabled. We were still able to display and translate blocks, menus, and views. Spanish Example: https://its.washcoopenhouses.org/es/table/intelligent-transportation-systems-open-house/conozca-al-equipo-de-sit

As for your config import error, this is because prior to adding in Spanish, the URL configurations in language's detection and selection settings was not properly set. When I added in Spanish without any changes to the modules, I got the following error in the Drupal UI: 

`Notice: Undefined index: source in Drupal\language\Plugin\LanguageNegotiation\LanguageNegotiationUrl->processOutbound() (line 142 of /home/ide/project/docroot/core/modules/language/src/Plugin/LanguageNegotiation/LanguageNegotiationUrl.php)`

Once that URL configuration for language detection was properly set, the error went away. So you will get an additional config import error in your bash only during the first import. Once the configuration is in the database, any imports following will no longer have that error.

PR #81, opened 2021-06-10
WCOR-20405: Configure a starter structure for group page menu views and breadcrumbs.

joshuami:
### Use Case

DigEx Team can see the basic page, resource, and group menu views and path alias settings that allow first draft of content to be added to County Drupal in a hierarchical structure.

**Implementation Details**

* Added field_states_ui and subpathauto to allow for menu UI and breadcrumb creation from new paths.
* Enabled help module to make it easier to find tokens (and solved the mystery of demonstrating block regions in the theme :o)
* Ordered fields for forms and default/teaser displays for the following:
** Groups
** Pages
** Resources
* Refactored resource description to use existing field_summary (missed that in the spec originally)
* Added help text for menu-related fields and turned on visibility controls for "show in menu" targeting sort weight (new field) and menu parent.
* Added entity reference view to restrict menu parent to pages or resources that do not have a parent already—and added path to help editor selection of correct content.
* Configured pathauto patterns to spec for the following:
** Alerts
** Contacts
** Groups
** Media (single pattern to rule them all)
** News
** Notifications
** Pages
** Resources
** Terms (default pattern for many terms)
** Users
* Set pathauto patterns to allow field_group_path as a safe token that can contain slashes.
* Added group_page_menu view that adds block menu views for group homepages, parent page showing children, and siblings of a common parent page. This view also has permission restricted management buttons that link to an admin view.

### Test Plan

`acli remote:drush @wcor.ode56 uli`

1. Add at least 2 pages to an existing group with "Show in menu" checked and no parent defined. If desired, make the sort weight difference override the short title alphabetical sort.
1. Add at least 2 more pages to the group with "Show in menu" checked and with one of the previous pages set as the menu parent.
1. Check that when viewing these pages a menu appears that allows you to see all top level pages from the group homepage, as well as the child pages of the parent page, and the siblings of a child with a link back to the parent.
1. Ensure that the deepest level page has breadcrumbs and a pathauto alias that includes its parent page.
1. Create at least one item of each of the types listed in pathauto patterns in the above implementation details. Ensure the paths make sense per the spec.


joshuami:
@lgbridges , this is building again with the small changes made during demo. It should be ready for your review and merge if you like the approach.

gatoona:
@joshuami okay to make push edits to this PR? Going to reenable interface translations and configuration translation in this PR and setup the correction detection order.

lgbridges (changes requested):
Code review passes. QA turned up one issue:

The Pathauto pattern "[term:parents:first]/[term:name]" used for taxonomy terms doesn't seem to generate the expected pattern according to spec tool of parent_taxonomy/term (e.g. news/news-type, projects/project-status-type). No parent taxonomy is included in the path--the term itself is right under the root.

joshuami:
> The Pathauto pattern "[term:parents:first]/[term:name]" used for taxonomy terms doesn't seem to generate the expected pattern according to spec tool of parent_taxonomy/term (e.g. news/news-type, projects/project-status-type). No parent taxonomy is included in the path--the term itself is right under the root.

This actually works as expected as I did not intend to put all the vocabulary patterns in place with this story. I wanted to talk through those a bit more with the team and see if we wanted to actually render out some of these taxonomy term pages as views or not. The "default" taxonomy term pattern is just to get something into place that will create a simple pattern for flat or hierarchical vocabularies. The parent part of the path is optional. It will basically create an alias just using the term name if the parent is not present. We don't have any hierarchical taxonomies just yet, so it just renders aliases with the term name off the root for our current list of vocabs.

We may have some cleanup to do on our terms on dev to get their path aliases corrected. I noticed that that terms created before we put in a pattern have their "generate automatic alias" checkbox unchecked. So just saving a term doesn't generate an alias. I can create a story for cleaning this up on dev and a follow up story to decide on the remaining vocabulary patterns after our brainstorming session on Friday.

I will go ahead and clean up the composer.lock conflict and resubmit this PR.

gatoona:
@joshuami There is a security update requirement for ctools that is failing pipelines. I've added it to #89

joshuami:
> @joshuami There is a security update requirement for ctools that is failing pipelines. I've added it to #89

Good to know. I'll need to rebase again once you've merged #89, which I just reviewed and approved. Please ping me when I'm good to rebase again.

lgbridges:
> Good to know. I'll need to rebase again once you've merged #89, which I just reviewed and approved. Please ping me when I'm good to rebase again.

@joshuami #89 has merged.

gatoona:
@joshuami @lgbridges another Acquia security requirement blocker for phpmailer. I'll update that and add to this PR since this is the earliest open PR we have now.

gatoona (approved):

lgbridges (approved):

PR #82, opened 2021-06-11
WCOR-0020525: Add and configure Views Bulk Ops & Edit.

jmilden-git:


lgbridges:
@jmilden-git Took a stab at filling out the story details including test plan -- please modify as needed.

**Use case**

As an admin, I would like to bulk update fields for selected content.

**Implementation**

Install and configure VBO. Adjust default system views to enable VBO Actions for selected items.

**Test plan**

As an administrator, open default system Media (/admin/content/media) and Content (/admin/content) views and attempt to apply all VBO actions to at least two items at a time.

lgbridges (changes requested):
Code review passes. QA issues are as follows:

All VBO actions work fine for items included in the default Media view (/admin/content/media). 

For the default Content view (/admin/content), VBO actions aren't working. For the _Modify field values_ action, many fields like Title, Short Title, Type, etc., are unable to be modified (the final _Apply_ button that would allow you to apply a change won't work for those fields). Other fields like _Displayed in_, however, were modifiable by VBO Modify Field Values. 

The VBO Publish/Unpublish actions for content also seem to have an incompatibility with Content Workflows, reporting an error "Page content items were skipped as they are under moderation and may not be directly unpublished." 

No immediate solutions were discovered.

joshuami:
@jmilden-git , I'm seeing the same stuff as @lgbridges on that CD environment.

It looks like we didn't test the publish/unpublish bulk operations during our chat. Those operations conflict with the content being under moderation. So we should at least remove those options from the bulk operations field on the admin content view.

The second oddity in this CD environment, is that clicking apply doesn't seem to even trigger the bulk operation on the second step of doing a bulk operation. We didn't see that happening on your environment during our call last week. There is some unapplied config (http://wcorode55.prod.acquia-sites.com/admin/config/development/configuration). I wonder if that could be impacting the functionality of the views bulk edit module. We'll need a clean build with all configuration imported to be sure.

Please try to rebase your branch on your local and drop me a note in Teams so we can take a look together on your branch and then try to push the change to the PR again.

jmilden-git:
Found a bug in Views Bulk Edit, so Josh will look at options later.  Closing PR, moving story to Backlog.

PR #87, opened 2021-06-15
WCOR-0: Changes to custom:catchup

gatoona:
Include pull:files, allow for command to be run in multiple dir inside project.


lgbridges (approved):
Tested in local IDE with no issues.

PR #89, opened 2021-06-16
WCOR-0: Fix workflow issue.

gatoona:
Fix workflow issue where setting moderated state to unpublish/publish did not reflect correct publish state of node.


gatoona:
@joshuami this PR fixes the workflow issue @jmilden-git brought up on today's sprint meeting.

config/default/workflows.workflow.editorial.yml
```diff
@@ -31,14 +31,13 @@ type_settings:
       weight: -1
     unpublished:
       published: false
-      default_revision: false
+      default_revision: true
```
joshuami, line 34: @gatoona, you beat me to it. I was just about to send a screenshot to the team chat that the issue was with the default revision setting 😄

config/default/workflows.workflow.editorial.yml
```diff
@@ -31,14 +31,13 @@ type_settings:
...
       label: Unpublished
       weight: 1
   transitions:
     draft:
       label: 'Create new draft'
       from:
-        - draft
```
joshuami, line 41: @gatoona, you should leave the settings that allow you to create a new revision in the same state as the current state. Creating 2-3 drafts before moving an edit to in review and have 2-3 in review comments or changes is pretty common. LIkewise, you'll want to be able to make an edit to published content and immediately publish that edit. 

Users get terribly confused when they don't see a moderation state for simple workflows like this. 😄

joshuami (changes requested):

config/default/workflows.workflow.editorial.yml
gatoona: @joshuami good point. Added back in. Did keep the transition from draft -> unpublish unchecked.

config/default/workflows.workflow.editorial.yml
```diff
@@ -65,7 +65,6 @@ type_settings:
     unpublish:
       label: Unpublish
       from:
-        - draft
```
joshuami, line 68: This is probably fine. I can't think of a great use case for unpublishing from any state other than published. 😄

joshuami (approved):

lgbridges (commented):
Is QA still needed?

gatoona:
> Is QA still needed?

@lgbridges feel free to merge.

PR #90, opened 2021-06-16
WCOR-0: Update README.md

lgbridges:
Spot edits to reflect the commit & PR naming convention change.


lgbridges:
The diff is busier than it should be because of a bunch of extra spacing that crept in.

joshuami (approved):
This looks solid to me. Adding an approval.

gatoona:
@lgbridges Don't know if you want to integrate #91 into the workflow. New command to check if pipelines will fail before deploying. Should save time.

PR #93, opened 2021-06-17
WCOR-0: Update core to 9.2.0

gatoona:
### ODE64

Changes made:

- Update core to 9.2.0
- Add in composer.json validation step to custom:check


docroot/.gitignore
```diff
@@ -10,4 +10,5 @@
 /index.php
 /robots.txt
 /update.php
-/web.config
\ No newline at end of file
+/web.config
+/README.md
```
gatoona, line 14: This was added in by core update. Does not ignore the root README.md file we have created.

lgbridges (approved):

PR #95, opened 2021-06-18
WCOR-0: Hotfix/webform libraries

gatoona:
### ODE 70

**Fix Case for #88** 

Fix Drupal status report warnings and mirror configurations from forms server.

**Implementation Details**

- Install webform libraries locally instead of relying on external CDN.
- Enable jquery_ui_datepicker.

**Test Plan**

- Check to see Webform library errors have been hidden in status report (/admin/reports/status).
- Successfully add a date picker field in a test form and collect its value when submitted.

PR #97, opened 2021-06-21
WCOR-1106: Enable and configure Acquia Search & Connector.

lgbridges:
## HOLD FOR QA

(Solr index/cores must be created for Dev and local environments post-PR before QA begins.)

**Use case**

As a site developer, I want to begin testing and optimization of search.

**Implementation details**

- Review all instructions from Acquia regarding Acquia Solr integration.
- Enable/configure module.
- Create the site connector.
- Configure the primary site search field.
- Test search and determine if any additional configurations are needed.

**Test plan**

As a site visitor (anon), I can search the site and receive any results (not necessarily accurate or optimized).


lgbridges:
I don't know why all the content types are suddenly appearing in the Pipelines "config doesn't match DB" error again, but I suppose I will try saving all form display modes and trying another push.

gatoona:
> I don't know why all the content types are suddenly appearing in the Pipelines "config doesn't match DB" error again, but I suppose I will try saving all form display modes and trying another push.

@lgbridges we need to merge #95 which includes a config export after the latest core update.

config/default/core.entity_form_display.media.image.default.yml
```diff
@@ -8,9 +8,7 @@ dependencies:
     - field.field.media.image.field_media_in_library
     - image.style.thumbnail
     - media.type.image
-    - workflows.workflow.media
```
joshuami, line 11: @lgbridges, this change seems a bit odd. We didn't remove images from being in a workflow, so I would expect this to still be a dependency. 🤔  Might be worth taking a look at the form and double checking that the image has a moderation field for published unpublished instead of just a publish state checkbox.

joshuami (changes requested):
Generally looks like a good enablement with a few small questions related to workflow changes that don't make much sense given what you would have been touching for search. It could just be a weird side effect from the scope of what search touches through enablement 🤔 

Also, this branch will need a rebase on top of the webform fixes. That will make the total files changed a lot smaller—though I kinda expected this to touch a lot of config because of search api excerpt and solr schema config files 😄

config/default/core.entity_form_display.media.image.default.yml
lgbridges: This was unexpected. I'll double-check it.

gatoona:
@lgbridges was able to get SOLR connected in #101 ODE80.

**Changes made:**

- Since we are running SOLR 3/4 and not 7, the following modules need to have strict versioning requirements:
```
"drupal/acquia_connector": "^3.0",
"drupal/acquia_search": "^2.0",
"drupal/search_api": "^1.0",
"drupal/search_api_solr": "^1.0",
"drupal/search_api_solr_multilingual": "^1.0",
```
- [Patch](https://www.drupal.org/files/issues/2021-03-07/2871721-for-2x-branch-34.patch) installed for acquia_search that fixes the search core connection issue. 
- Our account only has one search core. This is reflected in the [settings.php.](https://github.com/Washington-County/COUNTY-DRUPAL/pull/101/files#diff-b80b0a60af550f2090849fc9fd25fb09aa1022a91461c49e16fa1a2f0657ac54)
- Disabled sending site information to Acquia and ability to change site name. 
- Export correct search views page configuration.

gatoona:
Upon a deeper look, looks like we have both Solr 4 and Solr 7 on our account. Acquia's UI is just confusing as hell.
Successful builds are committed in #101 (Solr 4) and #103 (Solr 7) for you to review.

Some info I uncovered:

- If we use Solr 4, we only have one search core (JLQZ-203972) and that's applied at the entire application level.
- If we use Solr 7, each environment (ODE, Develop, Stage, Production) has its own search core (JLQZ-203972.prod.wcor, JLQZ-203972.dev.wcor, JLQZ-203972.ode42.wcor, etc)
- For Solr 7, [Acquia recommends using the default_drupal-4.1.1-solr-7.x-0 configset.](https://docs.acquia.com/acquia-search/managing-indexes/)
- The module support for Solr 4 seems to be going the way of the dinosaur so I suggest sticking with Solr 7.
- Using Solr 7, we should only be using JLQZ-203972.prod.wcor as the search core. I've gone ahead and set it up in the production environment. Other environments can connect to it if specified in the settings.php.
- Once we go live with the site, using config_split, we should make the index read-only for all environments except production so that content created in develop for example doesn't get indexed and show in production.

lgbridges:
Implementation unsuccessful due to inability to connect to Acquia Solr cores via acquia_search module 3.x. @gatoona was able to connect via acquia_search 2.x in test PR #103. I was unable to retrofit the changes from that PR into this PR, so, cancelling this PR and grooming #103 for the official merge.

PR #98, opened 2021-06-21
WCOR-1035: Configure summary field with maxlength.

jmilden-git:
### ODE75
**Use Case**

As an editor I need to enter the required summary on my content type and see how many characters are allowed and when I've reached the limit.

**Implementation Details**

* Should be required

* Should be 2 rows

* Should be used on all teaser views

* Should be maxlength 255

* Needs maxlenth RC1 module installed

**Test Plan**

For these content types (Group, Page, News, Resource) ensure that the summary field is required and maxlength is limited to 255 characters.


gatoona:
Taking QA and CR.

gatoona (approved):
Passes QA and CR

PR #102, opened 2021-06-24
WCOR-1089: Implement legacy paths and redirects.

lgbridges:
#### Use case

As an editor, I want to ensure requests for the CommonSpot / legacy site version of this page, group, or document are correctly handled/redirected to the current Drupal item.

#### Implementation details

Add Legacy path field to all relevant content types via [custom module code borrowed from eGovPDX repo](https://github.com/eGovPDX/portlandor/tree/master/web/modules/custom/portland/modules/portland_legacy_redirects).

#### Dependencies

- Redirect module

#### Test plan

As an editor, I can include the legacy path to enable redirection for pages, groups, and documents. I will only be allowed to enter a relative path (beginning with a slash, e.g. /HHS/AnimalServices/AnimalShelter/Adoption/index.cfm). I will not be allowed to enter a path with inappropriate characters, or to enter a legacy path more than once on the site.

As a site visitor (anon), my request to the CommonSpot page relative path will retrieve the current Drupal page. (In future story, www.co.washington.or.us requests will be routed to the new domain via separate redirect, preserving path.)


gatoona:
Taking QA and CR

docroot/modules/custom/wcor/modules/wcor_legacy_redirects/wcor_legacy_redirects.module
```diff
@@ -0,0 +1,41 @@
...
+    //       determine whether the field is enabled in form view?
+    $fields['field_redirects'] = BaseFieldDefinition::create('string')
+      ->setName('field_redirects')
+      ->setLabel(t('Legacy path'))
+      ->setComputed(TRUE)
+      ->setClass('\Drupal\wcor_legacy_redirects\Field\RedirectList')
+      ->setDisplayConfigurable('form', TRUE)
+      ->setDisplayOptions('form', [
+        'type' => 'text_textfield',
+        'weight' => 50,
```
gatoona, line 33: Since this adds the legacy field to every node entity, can we have the default for it be disabled so that we can select which content types the legacy field shows up in for form view? We could disable them manually, but seems like we have more entities that don't need it than those that do.

Implementing it would just require adding `'region' => 'hidden'` to the setDisplayOptions for form.

@joshuami any issues with this?

gatoona (changes requested):
Functions as intended.
Just a few minor changes/requests:

- [x] Add modules to core.extension.yml so they are automatically enabled on build.
- [x] Since this is adding a new field on entity form displays, we need to export the view configs for them with the field included. #104 includes a new custom:dex command that automatically does this once the module is enabled.
- [x] Disable the legacy field by default. (see code comment for wcor_legacy_redirects.module)

jmilden-git:
I'll take QA.

gatoona (approved):
@lgbridges everything looks good. QA and CR passes.

jmilden-git:
The files and media seem to be broken in the ODE82.  @gatoona how did you test the legacy path on those items?  Also, I thought we were disabling the field by default.

jmilden-git:
I added a redirect using the regular admin redirect field, but after saving see that it also appears as a "Legacy path" and looks available for an editor to modify.  Is that what we intended with this Legacy path field?

lgbridges:
> 
> 
> I added a redirect using the regular admin redirect field, but after saving see that it also appears as a "Legacy path" and looks available for an editor to modify. Is that what we intended with this Legacy path field?

I can't see any reason we wouldn't want an editor to be able to update the field. It will be their job to do this for non-MVP scope content. The field validation should keep them from incorrectly populating the field.

lgbridges:
> 
> 
> The files and media seem to be broken in the ODE82. @gatoona how did you test the legacy path on those items? Also, I thought we were disabling the field by default.

I'll check the ODE. The field should be hidden from form display by default. The only media type it should be active for is Document.

gatoona:
@lgbridges testing on my IDE, was able to get legacy paths to work on media entities like documents. Default core settings has standalone media URLs turned off. 

So if I set up a legacy path for a document it currently works like this:
/legacy-url.pdf -> /media/11/edit

If we turn it on it works like this:
/legacy-url.pdf -> /media/11

If we enable "Rabbit Hole media entity" and setup the correct rabbit hole configuration for media entities it then works like this:
/legacy-url.pdf -> https://awspath/file.pdf

This could be another story for setting up media paths.

jmilden-git:
> > The files and media seem to be broken in the ODE82. @gatoona how did you test the legacy path on those items? Also, I thought we were disabling the field by default.
> 
> I checked the ODE but I don't see anything broken. @jmilden-git -- can you tell me specifically what was broken?
> 
> The field should be hidden from form display by default. The only media type it should be active for is Document.

Yes, although it appears to be resolved, when first reported from the Content area any attempt to view Files or Media resulted in the "white screen" error.  No files or media could be tested at the time.

lgbridges:
> 
> This could be another story for setting up media paths.

Understood. I've made a note to draft this story on Monday.

jmilden-git (approved):

PR #105, opened 2021-06-25
WCOR-0: Solr update and configuration.

gatoona:
### ODE 84

**Use/Fix Case** 

- drupal/search_api_solr 4.1.12 is no longer supported and requires an immediate update to [4.2.0](https://www.drupal.org/project/search_api_solr/releases/4.2.0) 
- IDE/ODE setups can write to the production Solr index and can cause index pollution.

**Implementation Details**

- Update drupal/search_api_solr to 4.2 via composer.
- Enforce read-only for Solr indexing on IDE/ODE to prevent index pollution.

**Dependencies**

- drupal/search_api_solr
- drupal/acquia_search

**Test Plan**

- [x] Search API Solr in /admin/reports/updates is up to date and in version 4.2.0
- [x] In IDE or ODE, "Read only" is checked under Index options in /admin/config/search/search-api/index/acquia_search_index/edit


config/default/config_split.config_split.other.yml
```diff
@@ -10,6 +10,7 @@ module: {  }
 theme: {  }
 blacklist: {  }
 graylist:
+  - search_api.index.acquia_search_index
```
lgbridges, line 13: @gatoona -- Is this the equivalent of write protecting the index from "other" environments indexing? I'm assuming "graylist" just means the other envs can't access this config file, so index writes will fail?

gatoona: Naming convention is weird, but blacklist is for complete splits (will import no matter what), and graylist is for conditional splits (will only import if the configuration is different). So during the config import process, config split will check the graylist for the current active environment and only import configurations listed if they are different from the default. 

So in this example, config split will only import config/envs/ah_other/search_api.index.acquia_search_index.yml if it's different from config/default/search_api.index.acquia_search_index.yml and the current environment is an ODE or IDE.

gatoona: Also adding it under graylist ensures `drush cex` doesn't override the default sync file if a change is made on IDE to the search setting.

lgbridges (approved):

PR #106, opened 2021-06-25
WCOR-0: Clean up BLT custom commands.

gatoona:
### ODE 86

**Use/Fix Case** 

Clean up blt custom commands and outputs. Wiki documentation coming for guide on all `blt custom` commands.

**Dependencies**

- Acquia blt
- Acquia command line

**Test Plan**

- [x] `blt custom:pr [insert pr number]` will let you checkout a pull request on your environment.
- [x] `blt custom:ode:uli [insert ode number]` will output a ODE one time login link.
- [x] `blt custom:secret [insert ode number]` will upload the secrets.settings.php to the referenced CD environment.
- [x] `blt custom:secret` will download the secrets.settings.php from the develop environment to your IDE environment.


blt/src/Blt/Plugin/Commands/WashcoCommands.php
```diff
@@ -101,11 +106,13 @@ public function dex()
    */
   public function ode($cd)
```
lgbridges, line 107: One thing I was wondering was whether it would be a good idea to structure this command more like ``` blt custom:ode:uli ``` -- do we foresee any other kinds of actions on ODEs that would be convenient to handle via BLT?

lgbridges: Or ``` blt custom:ode-uli ``` if that's more straightforward to implement.

gatoona: Good idea. We might need to reserve the base 'ode' command in case we have additional ode commands in the future. I've made it custom:ode:uli instead of custom:ode-uli

lgbridges (approved):

PR #109, opened 2021-06-27
WCOR-0: Config split cleanup and fix.

gatoona:
### ODE 89

**Use/Fix Case** 

Appending to #100. In develop, stage, and production, the other environment is also active, causing split configurations in /config/envs/ah_other to also be imported. 

**Implementation Details**

- Assign a weight of -1 for other's configuration split settings.
- Use correct machine name, `ah_other`.
- Set default config for read-only indexing to true. 
- Set dev config for read-only indexing to false.
- Set default config for system.performance css/js aggregation to false.
- Set prod config for system.performance css/js aggregation to true.
- Disable configuration splits for ah_other, dev, stage, and prod. BLT automatically overwrites and activates the correct configuration split. 

**Dependencies**

- [drupal/config_split](https://www.drupal.org/project/config_split)
- Acquia blt

**Test Plan**

- [x] In an ODE/IDE, the value of "Authentication message"  in /admin/config/system/shield should be "Env: Other."
- [ ] After merge, in develop, the value of "Authentication message"  in /admin/config/system/shield should be "Env: Develop."
- [x] In an ODE/IDE, indexing for acquia_search_index should have read only enabled.
- [ ] After merge, in develop, indexing for acquia_search_index should have read only disabled.


config/default/config_split.config_split.ah_other.yml
```diff
@@ -1,18 +1,16 @@
...
-id: other
+id: ah_other
 label: Other
 description: 'Any Acquia environment not listed here (Cloud IDE and ODE).'
 folder: ../config/envs/ah_other
 module: {  }
 theme: {  }
 blacklist: {  }
 graylist:
-  - search_api.index.acquia_search_index
```
lgbridges, line 13: @gatoona -- Can we set IDE/ODE environments to also interact with Solr collections on a read-only basis?

lgbridges: Is this handled just by the fact that config split is disabled for IDE/ODE?

lgbridges (approved):
Will finalize QA post-merge.

PR #111, opened 2021-06-29
WCOR-1116: Ensure clean legacy redirect for documents.

gatoona:
### ODE 91

**Use/Fix Case** 

As a site visitor, I can request a document via a legacy path and be redirected to the canonical S3-hosted document URL (not the /media/[node-id] relative path).

**Implementation Details**

- Determine and implement correct config. Initial suggestion from Jung was to "enable Rabbit Hole media entity and setup the correct rabbit hole configuration for media entities using [media:field_media_file:entity:url] as the token for the rabbit hole redirect path.

**Dependencies**

- [drupal/s3fs](https://www.drupal.org/project/s3fs)
- legacy url module
- [drupal/media_entity_download](https://www.drupal.org/project/media_entity_download)
- [drupal/rabbit_hole](https://www.drupal.org/project/rabbit_hole)
- secrets.settings.php

**Test Plan**

- [ ] As a site visitor, requesting a legacy document directs me to the pathauto alias and shows me the document.
- [ ] As a site editor, requesting a legacy document directs me to the media's content page.


config/default/rabbit_hole.behavior_settings.media_type_document.yml
```diff
@@ -0,0 +1,14 @@
...
+status: true
+dependencies:
+  config:
+    - media.type.document
+id: media_type_document
+entity_type_id: media_type
+entity_id: document
+action: page_redirect
+allow_override: 0
+redirect: '[media:field_media_file:entity:url]'
```
joshuami, line 12: @gatoona, the only issue I see with using the redirect instead of https://www.drupal.org/project/media_entity_download is that your users will still see the S3 URL for PDFs and potentially bookmark that document. Media entity download actually delivers the media from the /media/[id] URL rather than showing the storage URL. That means that revisioned media will always have a permanent URL. Files with S3 are pointed to a specific revision in the files table. So each time you "replace" a file for a document or image, it creates a _0, _1, _2 version of the file. It will technically keep all those files unless you explicitly add a module or custom code to delete revisions. That means a bookmarked PDF could be a stale PDF. 😄

config/default/rabbit_hole.behavior_settings.media_type_image.yml
```diff
@@ -0,0 +1,14 @@
...
+status: true
+dependencies:
+  config:
+    - media.type.image
+id: media_type_image
+entity_type_id: media_type
+entity_id: image
+action: page_redirect
+allow_override: 0
+redirect: 'http://e0c8a7ed-3729-4812-8c88-ca7a002e115b.web.ahdev.cloud/media/11'
```
joshuami, line 12: I don't think you meant to redirect every image accessed by media URL to your local dev environment 😄

config/default/rabbit_hole.behavior_settings.media_type_video.yml
```diff
@@ -0,0 +1,14 @@
+uuid: dadd4985-a2f2-46f3-9ce8-8ef1482e0103
+langcode: en
+status: true
+dependencies:
+  config:
+    - media.type.video
+id: media_type_video
+entity_type_id: media_type
+entity_id: video
+action: display_page
```
joshuami, line 10: I would recommend this same setting for images. I can't think of a reason to redirect the image page.

joshuami (changes requested):
@gatoona , couple of comments and follow ups for you. It might be worth talking through the pros and cons of media_entity_download as a team. It may give you a little better user experience for documents with the slight drawback of costing a little more from a performance standpoint.

config/default/rabbit_hole.behavior_settings.media_type_video.yml
gatoona: I think you mean for videos? Implemented in latest commit.

config/default/rabbit_hole.behavior_settings.media_type_image.yml
gatoona: Haha. Damn auto input. Fixed in latest commit.

config/default/rabbit_hole.behavior_settings.media_type_document.yml
gatoona: May not be ideal since this forces the user to download the PDF and not be able to to open it directly in their browser.

joshuami: I double checked because not being able to set that seemed like a big miss for that module. 😄  It looks like you can set whether the media_entity_download link uses the attachment (download) or inline (in browser) content disposition header. (See https://www.drupal.org/project/media_entity_download/issues/3008834)

gatoona: Dope find @joshuami.

My latest commit includes media_entity_download and patch 35 for it. Looks like to have inline download, you have to append '?inline' to the end of '/media/xx/download' So I setup the rabbit hole redirect for documents to automatically go that route.

Now for site visitors requests go like this:
/media/xx -> /media/xx/download?inline (which opens the document file in browser)
/example-legacy-path.pdf -> /media/xx/download?inline

If a custom path alias is set, it works just as well.

Only weird issue is that /download?inline produced an error for media documents that were created prior to this implementation. Thankfully we only have one in test item in there.

joshuami: Getting close, but you don't actually need the rabbit hole for the media_entity_download implementation. 😄 

Documents that are embedded with media library will need a display mode that uses the download link widget for the file field on the media entity. You can see where this field is configured at `/admin/structure/media/manage/document/display`. 

2021-07-01 at 9 39 AM

To really make these documents accessible and usable when embedding, we can do a custom field template like: https://github.com/eGovPDX/employees/blob/master/web/themes/custom/cloudy2/templates/media/media--document--embedded.html.twig

And media library can support multiple display options when embedding, so we might want have an alternate output for the documents that includes a description: https://github.com/eGovPDX/employees/blob/master/web/themes/custom/cloudy2/templates/media/media--document--embedded-with-description.html.twig

gatoona: @joshuami if an author wanted to email out a direct link to open a PDF they uploaded in an email, I would prefer they send out a clean link that doesn't have /download?inline or /download?attachment in it.

I'll go ahead and enable the "Download link" format for the field label.
We'll need to revisit implementing a better output display once we have a custom theme in place to create/edit twig templates.

joshuami: @gatoona, good point about the link having a query parameter. It is not as clean, but it would definitely be compatible with the twig template. 🤔 

I did a little testing on your test document. I'm not sure why the automatic alias was turned off, and after turning it on, it produces a much more usable URL because of the subpathauto module. I also added a second test document to the BOC group here to test if the patterns were working correctly: https://wcorode91.prod.acquia-sites.com/boc/documents/2021-ap-style-updates/download?attachment

I noticed the link was attachment, so I took a look at the download link widget settings. I think you want to mark this to be browser default:

2021-07-01 at 10 15 AM

We might want to check the link for nofollow and open in new window as well. (Though I'm kinda torn on doing that even for documents because of the breaking the back link discussion we had yesterday.)

We'll want to be careful with the trim when we implement the twig template. I've seen cases where the trimmed output breaks the constructed link for the document.

gatoona: @joshuami dope suggestions. Latest commit sets the download behavior to "Browser default" and also enables autopath alias for documents so that it generates the path like your example. 

Link may not work now due to a new build with the commit, but I liked having https://wcorode91.prod.acquia-sites.com/boc/documents/2021-ap-style-updates/ redirect straight to the PDF for visitors, and having it redirect to the actual Drupal content page for logged in users. Could be useful to turn that view into a guide on how to properly share that document with a quick way for content authors to grab/copy the clean url to their clipboard. 

I'm on boat with not having the link open in a new window. One question, what would be the reason for enabling rel="nofollow" for document links?

Also, random test on mobile (Safari, iOS): Regardless of if it's ?attachment or ?inline, it treats it all as ?inline.

joshuami: That's a good point about redirecting the media page for anonymous users. If it gives folks a good link for sharing, maybe that's a good thing.

The main advantage to using `rel="nofollow"` is that it gives search engines a hint that we consider the link something that should not be individually indexed. This has some implications for content strategy and SEO. Getting Google, Bing, etc. to ignore the PDF can be helpful when you want people to find the page containing the PDF. It's not a guarantee that Google won't index the PDF, but it reduces the likelihood.

Interesting that Safari just ignores the content-disposition header. I wonder if it does it for all downloadable file types or just PDF because of the nature of PDFs. We could always do a quick Word or Excel test to see if it presents download in those cases. 

We should probably test those other document types for Chrome, Edge, Firefox, etc. just so that we know the behavior for each. I feel like Firefox used to default to attachment and force a download regardless of the content-disposition header. I don't know if that is still the case. Sharing that information with the team so that it can make its way into training will be important as well. The more you know 🌟

lgbridges (commented):
@gatoona -- Can you expand on the test plan a bit? I'm not sure if I'm triggering what we're after. I'm retrieving documents via /documents/[document-title]/download?inline. Not sure if that's right.

gatoona:
> @gatoona -- Can you expand on the test plan a bit? I'm not sure if I'm triggering what we're after. I'm retrieving documents via /documents/[document-title]/download?inline. Not sure if that's right.

@lgbridges Are you checking the link as a visitor (logged off).

lgbridges (approved):

joshuami (approved):

lgbridges (approved):

PR #114, opened 2021-07-01
WCOR-1110: Add reCaptcha module.

jmilden-git:
### ODE 93

**Use case**

As a site admin, I want to enable reCAPTCHA for all public-facing forms.

**Implementation details**
Install and configure reCAPTCHA.

Dependencies
CAPTCHA module (already installed)

**Test plan**

As a site admin, I can place the reCAPTCHA field on a new webform.
As a site visitor (anon), I must pass the reCAPTCHA challenge to successfully submit a webform.


config/default/recaptcha.settings.yml
```diff
@@ -0,0 +1,12 @@
+site_key: 6LfigcYUAAAAAMppl36pM4mw8FFmNe0kGtV9hnZ5
+secret_key: 6LfigcYUAAAAADX6ARiY21OWi9zs7pMjW2ydMcb7
```
gatoona, line 2: Hey @jmilden-git, we should not be committing secret keys/credentials into the public repo.

gatoona (changes requested):
Please see comments relating to code.

Pending tasks:

- [ ]  Remove site_key and secret_key from commit.
- [ ] Generate the site key under the webteam's group Google account.
- [ ] Insert site_key and secret_key into secrets.settings.php

config/default/recaptcha.settings.yml
jmilden-git: OK, those keys came from our existing reCaptcha Webteam account.  Is there another account I'm missing?  I don't see the secrets file in the repo.  Is this documented somewhere?

gatoona: @jmilden-git take a look at https://docs.acquia.com/resource/secrets/

Also, looks like the site and secret key you used was for another site application and you added the ACP domains into that. Please remove them from that site and create a separate site with new site/secret key to implement.

I will have to go and delete this site now that the keys have already been compromised.

gatoona (approved):

PR #115, opened 2021-07-07
WCOR-0: Reinstall packages after being removed prior to a patch install.

gatoona:
### ODE XX

**Use/Fix Case** 

When adding/removing a patch to a package that has already been installed, composer removes the package but does not reinstall the package back during the initial run of composer install.

**Implementation Details**

- Run `composer update --lock` right after `composer install`

**Test Plan**

- [x] When adding/removing a patch from composer.json `blt custom:catchup`  removed packages should be installed back.


joshuami:
@gatoona , I think this might help get the lock file updated when there was an incorrectly resolved git merge—like when the second story in a row touches composer.lock and you get an out of date hash—but it shouldn't do anything in regards to patching. Patches are only affected by composer.json, they are not written to composer.lock. Whether you do `composer update` or `composer install` you should get your patches based on what is currently present in the composer.json patches section.

joshuami (approved):

gatoona:
@joshuami I didn't want to run `composer install` twice, but `composer update --lock` does seem to reinstall the package and install the patch into the /vendor directory after it has been removed from the first composer install run.

You can replicate the steps with the latest develop branch:

1. Remove/add the csv_importer patch in composer.json
2. Run `composer install`
3. You will see that csv_importer has been uninstalled and not been reinstalled.
4. Run `composer install` again or `composer update --lock` and the package/patch will be reinstalled.

Not sure if it's a bug with composer in general.

gatoona:
@joshuami scratch that last comment. Looks like a patch for it is out https://github.com/cweagans/composer-patches/releases/tag/1.7.1

Will apply this and test it out.

joshuami:
> @joshuami scratch that last comment. Looks like a new release for it is out https://github.com/cweagans/composer-patches/releases/tag/1.7.1 that resolves the issue.
> 
> Will apply this and test it out.

Great find. I have not run into the issue yet, but I'm betting I will if I upgrade a project to 2.1. 😄

lgbridges (approved):

PR #116, opened 2021-07-07
WCOR-1050: Identify and add language set for core multilingual.

sharstine:
**ODE 95**

**Use case**

As an editor, I want to be able to enter translated website content in the County's safe harbor languages.

**Implementation details**

1. Confirm the County's safe harbor languages with Philip. (Update: Philip confirmed the list below.)

   - Arabic
   - Chinese (Cantonese & Mandarin)
   - Filipino/Tagalog
   - Japanese
   - Khmer/Cambodian
   - Korean
   - Persian
   - Russian
   - Somali
   - Spanish
   - Vietnamese

2. Add language names and language codes in Drupal via /admin/config/regional/language.

**Test plan**

1. Create a new page in Drupal.
2. Click the "Translate" tab on the node once it's saved.
3. Verify that English and all the County's safe harbor languages are listed.
4. Click "Add" to the far right of any language to enter a translation of the English page you created.
5. Replace English content with sample translated content in the required fields.
6. Click "Save (all translations)."
7. Verify that the new translated page gets a path beginning with the language code. For example, if the English URL is /administration-hhs-directors-office/test, the Spanish URL would begin with /es/ and might be something like /es/administration-hhs-directors-office/prueba.


sharstine:
QA can begin on ODE95.

lgbridges (changes requested):
Code review passes. QA uncovered some issues. Additional testing may be needed to verify (perhaps I took a wrong turn). We could work out the first bullet in a follow-up story, but I think the pathing issue has to be addressed in the scope of this story.

- Creation of the translated page generates an *additional* new item in the general content admin page (/admin/content), but there is no column showing the language version (basically, showing that it is a translation). We can obviously create a new view from scratch that we can use to administer content, but it doesn't seem like we should *have* to do that. Is this a known issue for core multilingual? Is there evidence of how others are addressing it, or working around it?

- The last item in the test plan requests verification that the relative path starts with /[language-code]. Noticed that the **page title portion** of the example Spanish URL there was translated. I believe this is not an automated behavior--meaning you would have had to elect to update URL alias manually on that translation. If you change the URL alias on the English page, everything is fine. But if you change the URL alias **on the translated page**, (A) the alias entered on the translated page becomes the **canonical / go-to** alias for the English page **and** all others, (B) the URL scheme for the translated page no longer works as expected, meaning that you can't access the translated page via /[lang]/[group]/[new-alias], only /[lang]/[group]/[original-alias]. It could be that adjustments to Pathauto patterns may address this issue. There may need to be patterns established for translated content types, one pattern for English (without the initial lang code path level) and one for non-English. This might be enough to address the creation of new aliases after the initial translation is created.

gatoona:
@lgbridges 

> Creation of the translated page generates an additional new item in the general content admin page (/admin/content), but there is no column showing the language version (basically, showing that it is a translation). We can obviously create a new view from scratch that we can use to administer content, but it doesn't seem like we should have to do that. Is this a known issue for core multilingual? Is there evidence of how others are addressing it, or working around it?

We would just add a translation language field to the existing content page view. /admin/structure/views/view/content
Take a look at washcoopenhouses.org and how it manages translated pages for admin/content.
We should probably also add the ID field so that we can see that they are the same content, but just translated. We can even go as far to group by ID in the view so that all the translations for a single content are shown together.

gatoona:
@lgbridges @sharstine 

> The last item in the test plan requests verification that the relative path starts with /[language-code]. Noticed that the page title portion of the example Spanish URL there was translated. I believe this is not an automated behavior--meaning you would have had to elect to update URL alias manually on that translation. If you change the URL alias on the English page, everything is fine. But if you change the URL alias on the translated page, (A) the alias entered on the translated page becomes the canonical / go-to alias for the English page and all others, (B) the URL scheme for the translated page no longer works as expected, meaning that you can't access the translated page via /[lang]/[group]/[new-alias]/[lang]/[group]/[new-alias], only /[lang]/[group]/[original-alias] /[lang]/[group]/[original-alias]. It could be that adjustments to Pathauto patterns may address this issue. There may need to be patterns established for translated content types, one pattern for English (without the initial lang code path level) and one for non-English. This might be enough to address the creation of new aliases after the initial translation is created.

For this issue, we simple have to select, "Do nothing. Leave the old alias intact." under "Update action" at /admin/config/search/path/settings

This should make it so that if the title to the Spanish version changes, the old English URL still remains for it and it can be accessed via just appending /es to the URL.

lgbridges:
> 
> For this issue, we simple have to select, "Do nothing. Leave the old alias intact." under "Update action" at /admin/config/search/path/settings
> 
> This should make it so that if the title to the Spanish version changes, the old English URL still remains for it and it can be accessed via just appending /es to the URL.
> 
> In addition to the above, we will also need to use [node:source:title] to set the path alias instead of [node:title]

Agreed, and glad you caught this. This is how I have traditionally set up Pathauto, so I definitely prefer this. However, this will impact the site more broadly, so I'm not sure if we need to get team consensus.

lgbridges:
> 
> We would just add a translation language field to the existing content page view. /admin/structure/views/view/content
> Take a look at washcoopenhouses.org and how it manages translated pages for admin/content.
> We should probably also add the ID field so that we can see that they are the same content, but just translated. We can even go as far to group by ID in the view so that all the translations for a single content are shown together.

I think this will work. This could spin off into a new story.

gatoona:
@lgbridges unless there's any objection, I'll go ahead and make those changes for @sharstine and contribute to this PR.

lgbridges:
> 
> 
> @lgbridges unless there's any objection, I'll go ahead and make those changes for @sharstine and contribute to this PR.

@sharstine, @gatoona  -- the Pathauto pattern adjustments didn't specifically correct the mentioned issue. However, I noticed that URL aliases were not checked for Page or News types in the Content language config page (/admin/config/regional/content-language). Unless I hear an objection, I'm going to push a commit here that enables URL alias translation for all content types currently allowing translation.

gatoona:
> the Pathauto pattern adjustments didn't specifically correct the mentioned issue.

@lgbridges are you referring to the issue where the title portion of the URL changes for translated nodes? It was resolved for me for news types.

> However, I noticed that URL aliases were not checked for Page or News types in the Content language config page

Do we want URLs to be translated for translated nodes? My assumption was that we want the URL to be the same with the exception of the language code being placed in to switch.

lgbridges:
> @lgbridges are you referring to the issue where the title portion of the URL changes for translated nodes? It was resolved for me for news types.

I'll re-test this morning.
 
> Do we want URLs to be translated for translated nodes? My assumption was that we want the URL to be the same with the exception of the language code being placed in to switch.

I've read conflicting commentary on best practices here. From an SEO perspective, we're probably covered if we have the language path prefix in the URL before the page title. Less clear from a usability standpoint what it means to have an English title in a URL being accessed by non-English speakers. I'm willing to keep it simple for now and turn URL aliasing off for translation, as long as the basic functionality stands up.

lgbridges (approved):
I am passing this for CR & QA. 

There is an issue occurring when URL alias _changes_, but this is not in the scope of the use case or test plan and needs to be its own correction story. The current behavior for paths, provided URL alias is not changed, is as expected: translations are prefixed with a _/[lang-code/_ path level.

PR #117, opened 2021-07-07
WCOR-1061: Implement map and transit links for buildings.

lgbridges:
### ODE 96

**Use case**

As a site visitor, I am able to click a link on a given Building item to open a Google Map showing the location according to the lat/long coordinates in the Geolocation field. Another link is similarly presented for TriMet Trip Planner.

**Implementation details**

Josh suggested the use of Entity Views Attachment to address the need to render the link dynamically as part of node display without storing a field value. 

**Dependencies**

- EVA (Entity Views Attachment)

**Test plan**

As an editor, I am able to update the Geolocation field of Building content and see the correct Google Maps link and TriMet Trip Planner link displayed upon cache clear.

As a site visitor (anon), I am able to view the Google Maps link for every Building (since Geolocation is a required field) and see an accurate map pin assigned to the "To:" field of the map directions tab. I'm also able to open TriMet Trip Planner with the destination address preopulated.


gatoona:
Taking QA and CR.

gatoona:
Some feedback/questions on QA:

1. I was under the impression we would not be forcing new windows opening on users. https://knowbility.org/blog/2019/links-opening-new-windows-no-warning
2. Is the new "Get Directions" link the solution for the transit information checkbox? If not, where would the transit information link be displayed in addition to this link? Maybe we just remove the transit information checkbox then?

lgbridges:
>     1. I was under the impression we would not be forcing new windows opening on users. https://knowbility.org/blog/2019/links-opening-new-windows-no-warning

It was a point made by Josh regarding accessibility in one of our meetings. I couldn't remember if we had a consensus about it then, or earlier. I am somewhere between neutral and leaning toward same tab. I am happy to change it to opening same tab for now and we can discuss further in retro/demo.

>     2. Is the new "Get Directions" link the solution for the transit information checkbox? If not, where would the transit information link be displayed in addition to this link? Maybe we just remove the transit information checkbox then?

It wouldn't be a bad idea to deploy the transit link in the same way (but separately). I will take a second look today to see if that's possible.

jmilden-git:
@gatoona @lgbridges I think the important point is not to do it without warning.  So when the link points to a site other than our own it should be indicated that they are leaving or the link opens in a new window.  It's mainly a usability issue.

lgbridges:
> 
> 
> @gatoona @lgbridges I think the important point is not to do it without warning. So when the link points to a site other than our own it should be indicated that they are leaving or the link opens in a new window. It's mainly a usability issue.

@jmilden-git -- The implementation path isn't clear. I'm thinking of the many mobile users (our primary audience at this point) who will end up being directly sent to the Google Maps app on their phone when they click this link. Trying to envision the mechanism for pausing to convey the information. Something like a splash screen that pauses before switching to the app or loading the web version of GMaps into the tab? I'm struggling to think of an example--if you have one to share, that would be helpful.

gatoona (approved):

PR #118, opened 2021-07-08
WCOR-1119: Create group reference view and apply to content types.

gatoona:
### ODE 97

**Use/Fix Case** 

When selecting a group (department, division) to place a content under, there are multiple divisions with the same title (Ex. administration) making it hard to distinguish which department the division is under. As an editor, I would like to be able to tell the difference between divisions with the same title.

**Implementation Details**

- Create reference entity view for Groups and include parent group values.
- Configure reference entity autocomplete widget to use the view created.

**Test Plan**

- [x] Create a test content like news and type in "administration" in "Displayed in." You should be able to see all the administration divisions and which department it belongs under.


jmilden-git:
I can take QA for this one.

jmilden-git:
QA passed.

jmilden-git (approved):

PR #122, opened 2021-07-15
WCOR-1070 editor and publisher roles and views for my groups, my content, and my media.

joshuami:
### ODE 102

**Use/Fix Case** 

As an editor, I need to be able to create pages, news, and put them into an in review workflow state; and the editor should be able to create images, documents, and video and put it into a published state for embedding into other content.

As a publisher, I need to be able to to all of the above and also move content into published.

As an editor or publisher, I should be able to see content in my groups for a personal dashboard.

As a site builder, I need to be able to masquerade as an editor or publisher and have that action logged in the dblog.

**Implementation Details**

1. Enabled masquerade, masquerade_field, and masquerade log. Add field to user profile and set permission to only administrator having access to the feature.
2. Created editor and publisher role, and configured permissions per the use case.
3. Added a a "groups" field to the user entity for tracking which groups you expect a user to do edits.
4. Used field permissions so that groups can only be edited by site administrators.
5. Created "my content", "my media", and "my groups" admin page and add them to a toolbar menu.
6. Updated the default short date format to use time as well so that administrative views can better show when an item was changed.
7. Removed the administrative menu functions from users with just the editor or admin role.
8. Put "Add content" and "Add media" in the editor toolbar.
9. When configuring images to be related to groups (necessary for my media view), I noticed some missing fields for synced fields and incorrectly configured translation settings. Exporting after those fixes created quite a bit of file change. :) 
10. Added the "marked for deletion" workflow state and configuration permissions for publisher to use.

**Dependencies**

- Couple of new modules added: field_permissions, masquerade, masquerade_field, masquerade_log, and toolbar_menu.
- There may be some merge conflicts with stories that are also adding modules and/or with Joe's responsive images and media library story.

**Test Plan**

- [x] As an administrator, create a new user "Polly Publisher" with the role *publisher* and assign them to a couple of groups using the groups field on the user profile. Use a WashCo email for this user that will not be related to any real user or inbox. (Do not use an `@test.com` or `@example.com` address for security reasons.) 
- [x] As an administrator, create a new user "Eddie Editor" with the role *editor* and assign them to a couple of groups using the groups field on the user profile. I recommend making at least one of the groups match to Polly Publisher's list of groups.
- [x] As an administrator, view Eddie Editor's profile and use the "masquerade as" link to see the site as Eddie.
- [x] As Eddie Editor, create some content in the same group as Polly. Test at least a couple of content types. Eddie should be able to create content, but will not be able to publish the content.
- [x] As Eddie Editor, create some media in the same group as Polly. Test at least a couple of media types, preferably all three. Eddie should be able to create media and it should be published by default.
- [x] Use the "unmasquerade" link to become an administrator again. 
- [x] As an administrator, view Polly Publisher's profile and use the "masquerade as" link to see the site as Polly.
- [x] As Polly Publisher, view your "My groups", "My content", and "My media" views.
- [x] As Polly, update content in these groups and move at least a couple to "published".
- [x] Polly should be able to edit and publish just about everything, but the "my" views should limit their visibility of content that is not their responsibility.
- [x] Add menu items for "Add content" and "Add media" to the editor toolbar menue

**Post Deployment**

- [ ] You'll need to create Polly and Eddie on the dev environment after this story has merged. That will make them available for testing in future stories.
- [ ] Add menu items for "Add content" and "Add media" to the editor toolbar menue

**Followups**

Need to decide the best approach to media workflow. It's currently configured, but not hooked up to any media types in favor of a simple published/unpublished field. Since non-admins can't delete media, it stands that we might end up with some duplication without a way to see what should be removed.


jmilden-git:
I'll take QA for this one.

jmilden-git (dismissed):

composer.json
```diff
@@ -115,14 +120,12 @@
...
     "patches": {
         "drupal/core": {
-            "2719797 - New option for Views page displays to use the admin theme - feature/STRY0020001": "https://www.drupal.org/files/issues/2020-10-07/2719797-109.patch"
+            "2719797 - New option for Views page displays to use the admin theme - feature/STRY0020001": "https://www.drupal.org/files/issues/2020-10-07/2719797-109.patch",
+            "3129874 - The 'moderation_state' base field overrides cause install from existing config to fail": "https://www.drupal.org/files/issues/2021-02-08/fix_moderation_state_bfo_dependencies_approach6-3129874-24.patch"
         },
         "drupal/scheduler_content_moderation_integration": {
             "3077147 - Scheduler content moderation integration error fix - feature/STRY0020225": "https://www.drupal.org/files/issues/2020-06-30/3077147-28.patch"
         },
-        "drupal/media_entity_download": {
```
gatoona, line 123: @joshuami reason for this being removed?

joshuami: I'm glad you noticed 😄 

To get a clean composer.lock after fixing merge conflicts, I opted for a `composer update` as `composer update --lock` wasn't doing the trick. That gave us the following updates:

```
  - Upgrading aws/aws-sdk-php (3.185.14 => 3.185.17)
  - Upgrading doctrine/cache (1.12.0 => 1.12.1)
  - Upgrading drupal/geofield (1.32.0 => 1.33.0)
  - Upgrading drupal/media_entity_download (2.0.0 => 2.1.0)
  - Upgrading drupal/s3fs (3.0.0-beta1 => 3.0.0-beta2)
  - Upgrading drupal/scheduler (1.3.0 => 1.4.0)
  - Upgrading drupal/search_api (1.19.0 => 1.20.0)
  - Upgrading http-interop/http-factory-guzzle (1.0.0 => 1.1.0)
  - Upgrading phpunit/php-file-iterator (2.0.3 => 2.0.4)
  ```
The new media_entity_download version included the patch, so it could be removed.

We should rerun the QA steps against the ODE environment as that should give us pretty decent test coverage for these updates.

lgbridges:
@joshuami @gatoona -- Just checking in on what's left to do here. Is additional QA needed or just final review and merge?

joshuami:
> @joshuami @gatoona -- Just checking in on what's left to do here. Is additional QA needed or just final review and merge?

Someone should walk through the test plan one more time to confirm that the included composer update did not affect anything. Then a final review would be warranted.

If no one gets to the QA plan before demos this afternoon, I can do the test plan as the demo on the ODE environment.

joshuami:
@lgbridges , I think this is good to go no. No more merge conflicts and even squeezed in a fix to the moderated content view that will allow the DigX team to process content marked for deletion 😄 

Please feel free to give it a final review and merge.

lgbridges (approved):

PR #125, opened 2021-07-20
WCOR-1082: Add social media fields to Group content type.

sharstine:
### ODE 104

**Use case**
As a community member, I want to be able to follow a workgroup's social media.

**Implementation details**
- Add social media fields (Facebook, Twitter, YouTube, and LinkedIn) to the Group edit form per Spec Tool.
- Output social media fields to the Group homepage.

**Follow-up stories (added to Story Brainstorm spreadsheet)**
- Work with design firm to determine theme needs.
- Add custom handling/validation to domains entered into the social media fields.
- Customize domains in default help text, possibly using hook_form_alter.

**Test plan**
As an admin, I can add social media links to a new or existing Group and see the output of those fields on the Group homepage.


lgbridges (approved):
CR / QA passes. Noting the follow-up stories note (thank you for adding them)--it will be good to see some display refinement and input validation at a later point.

**Suggestion:** Presumably, external link fields like these shouldn't be set translatable--the Spec Tool was incorrect. This can be corrected in this sprint, or as part of a larger "translatable field cleanup" story later on.

PR #127, opened 2021-07-21
WCOR-1121: Define and implement contact fields for groups.

lgbridges:
### ODE 110

**Use case**

Contact fields are fully implemented for groups sufficient for final import from Central System dataset.

**Implementation details**

Field definitions should be entered into the Spec Tool. The team's discussion led us to the following conclusions:

- Email should be standard and listed first. 
- Phone should be standard and listed second. The format should be xxx-xxx-xxxx and the link href prefix should be "tel:" for mobile integration. (Josh has suggested borrowing code from POWR: ([https://github.com/eGovPDX/portlandor/tree/master/web/modules/custom/cleave]https://github.com/eGovPDX/portlandor/tree/master/web/modules/custom/cleave))
- Fax numbers **will not** be listed at the group level, only in individual pages as part of a described workflow.
- Social media links will be provided from a separate story. (Not required for the final group data import from Central System dataset.)
- A given group should inherit contact details from its nearest parent group containing contact details, unless the group itself wishes to define its own contact details (or had contact details in Central System that have been imported). This is likely the scope of a separate story that creates conditional views for contact block display, etc.

**Test plan**

- [x] Add a group, making sure to populate Email address and Phone fields.
- [x] Ensure the email address and phone number are displayed and correctly linked. (Phone link target should be prefixed by "tel:" to invoke a phone app.)


lgbridges:
Regarding the phone extension add-on discussed in sprint retro, I will spin off a new story to address that.

gatoona (approved):

PR #128, opened 2021-07-21
WCOR-1079: Migrate all SO media releases prior to 2021

gatoona:
### ODE 111

**Use/Fix Case** 

Import SO media releases.

**Implementation Details**

- Install and enable feeds module for automating file uploads.
- Create new file field in news for PDFs.
- Configure rabbithole for news so that if a PDF exists for a news item, it will redirect to that PDF.

**Dependencies**

- [drupal/feeds](https://www.drupal.org/project/feeds)

**Test Plan**

- [x] Create a news item and upload a PDF for it. Visit the news page logged off. It should redirect you to the PDF instead of the node's page view.
- [ ] Importing of archive media releases to be continued in dev.


lgbridges (approved):
CR / QA passes.

**Suggestion:** Now or prior to launch, set field-specific permissions on field_pdf_news so that only admins can update. PDF substitutions for news pages is strongly nonpreferred -- we would only want to keep this field editable if a PDF News workflow was approved by CAO / Steering Committee.

**Suggestion:** *If this is approved to be a writable field by authors*, set a size limit for PDFs. Generally, only very long PDFs or complex map PDFs, etc, are prone to going over 256MB, and those shouldn't be in general news. 

**Question:** Will there be a follow-up story to remove Feeds after the migration?

PR #136, opened 2021-08-03
WCOR-1138: Install and configure Linkit module.

gatoona:
### ODE 118

**Use/Fix Case** 

[insert here]

**Implementation Details**

- Install and enable linkit module.
- Configure "Basic HTML" text format to setup linkit.
- Configure linkit to show group value in results.

**Dependencies**

- [drupal/linkit](https://www.drupal.org/project/linkit)

**Test Plan**

- [x] In a HTML field (body), use the link tool to search for and insert published pages. In source view the href value will be node/*, but when viewed outside ckeditor, the link automatically uses the proper url alias generated from pathauto.


jmilden-git (dismissed):
QA passes.

config/default/linkit.linkit_profile.default.yml
```diff
@@ -0,0 +1,25 @@
...
+description: 'A default Linkit profile'
+matchers:
+  556010a3-e317-48b3-b4ed-854c10f4b950:
+    uuid: 556010a3-e317-48b3-b4ed-854c10f4b950
+    id: 'entity:node'
+    weight: 0
+    settings:
+      metadata: 'by [node:author] | [node:field_group] | [node:created:medium]'
+      bundles:
+        page: page
```
joshuami, line 20: @gatoona , any reason you limited this to just pages and resources? It seems like someone might link to news or groups as well.

config/default/image.style.linkit_result_thumbnail.yml
```diff
@@ -0,0 +1,17 @@
+uuid: 1540514c-f2a1-48b4-a423-2eca02f12fa6
+langcode: en
+status: true
+dependencies: {  }
+_core:
+  default_config_hash: CMuvuRCfGZ5k3XgsXmkEc54rI8etDY1Qz2JP7JzKL-g
+name: linkit_result_thumbnail
```
joshuami, line 7: I'm curious about this image style. I'm assuming it was added by the module, so it doesn't seem to be in use. Not anything that should fail the PR, but definitely interesting to look into.

config/default/linkit.linkit_profile.default.yml
```diff
@@ -0,0 +1,25 @@
...
+        resource: resource
+      group_by_bundle: true
+      include_unpublished: false
```
joshuami, line 23: If you want your editors to be able to stage a set of pages to be published at once, you may want to include unpublished.

joshuami (approved):
Nothing fails the PR, but it would be worth reviewing the comments.

config/default/linkit.linkit_profile.default.yml
gatoona: True. My concern here was from authors linking to unpublished content on a published content.
I can see this happening a lot in a moderated workflow.

config/default/linkit.linkit_profile.default.yml
gatoona: Included news and groups as a separate matcher so it shows below initial page/resource results.
Initial issue was because you couldn't reorder news results to be below page results on a single matcher.

lgbridges (approved):
CR passes; QA re-confirmed.

PR #137, opened 2021-08-03
WCOR-1127: Simple custom theme to start twig development.

gatoona:
### ODE 119

**Use/Fix Case** 

As a developer, I would like to create custom twig templates within a custom theme.

**Implementation Details**

- Create basic custom theme using Bootstrap 4.

**Dependencies**

- Bootstrap 4

**Test Plan**

- [x] Site loads without error.
- [x] The custom theme, "WashCo" is enabled and set as the default.


docroot/themes/washco/templates/page.html.twig
```diff
@@ -0,0 +1,23 @@
+<div class="container" id="content-section">
```
joshuami, line 1: I would restructure this as follows to give us a mobile and accessibility friendly layout from the start:

```
<div class="container" id="content-section">
  <div class="row">
    {% set has_sidebar_first = page.sidebar|render|striptags|trim is not empty %}
    {% if has_sidebar_first %}
      <div class="main-col content-col order-1 order-md-2 col-12 col-md-10 margin-y">
        {% if label %}
        <h1>{{ label }}</h1>
        {% endif %}
        {{ page.content }}
      </div>
      <div class="side-col order-2 order-md-1 col-12 col-md-2 margin-y">
        {{ page.sidebar }}
      </div>
    {% else %}
      <div class="col-12 main-col content-col margin-y">
        {% if label %}
        <h1>{{ label }}</h1>
        {% endif %}
        {{ page.content }}
      </div>
    {% endif %}
  </div>
</div>
```

joshuami (changes requested):
A few small recommendations and clean ups. Out of curiosity, did this get generated from a starter theme or was it more pieced together? I was originally thinking we could generate a base theme from Radix or similar to get the necessary templates for a navbar as well as the structure for doing a SaSS build out using commands like `npm watch`. That's not saying we can't add that as a follow up after hashing out a theming plan with IF. 😄

gatoona:
@joshuami changes made and committed. 

> Out of curiosity, did this get generated from a starter theme or was it more pieced together?

This was pieced together from scratch to keep the theme as simple as possible and to keep the focus solely on twig development for content. Not that we couldn't do that with an existing starter theme, but it's easier IMO to start off with a clean slate that doesn't already have a ton of existing twig files, especially if down the road we might need to merge our twig files with IF's packaged theme.

lgbridges (dismissed):
@gatoona -- CR & QA passes.

I don't think any questions about this theme becoming our new default arose during your demo, and I know you and Josh are volleying a bit here about implementation.

This theme is obviously not meant to represent the future site design in any real aesthetic/design way, but I have gotten a bit used to the layout we'd been working with in Olivero in so far as our sidebar is on the right. I tend to prefer that layout at this point, and I'm wondering if Sidebar could be swapped with Content region so Content is left and Sidebar is right. However, I'm only one person on the team and to my knowledge we haven't landed on team consensus about standard layout, much less gotten I.F. into that conversation. Curious about your thoughts. Could be a separate story.

gatoona:
@lgbridges the sidebar has been moved to the right. Latest commit also enables twig debugging in IDE.

Thanks!

lgbridges (approved):

PR #140, opened 2021-08-06
hotfix/solr-dev: Disable read-only for all environments. Sync solr da…

gatoona:
…tabases to correct environments.

### ODE 122

**Use/Fix Case** 

As an admin, I would like to test against solr databases without affecting production.

**Implementation Details**

- Build Solr 7 databases for test and dev.
- Configure IDE/ODE/DEV to use dev instance of Solr. Stage to test. Prod to prod. 
- Disable read-only and enable 'index items immediately,' for all environmnents.

**Dependencies**

- Solr 7
- Acquia Search API

**Test Plan**

- [x] Ensure Solr connection can be reached and is pointing to the DEV Solr instance for ide/ode/dev: /admin/config/search/search-api/server/acquia_search_server
- [x] Ensure indexing is working: /admin/config/search/search-api/index/acquia_search_index


joshuami (approved):
Changes look good.

PR #142, opened 2021-08-11
WCOR-1135: Enable facets module and configure default search view

lgbridges:
**Use case**

As a community member, I want to be able to narrow my search results by type and department.

**Implementation details**

Install the search api facets module and configure two facets for the default site search. Place the resulting facet blocks in the side region for this view using a page limit (by path) in the block layout. You will probably need to add the field to the search api index to make it available for creating a facet.

**Test plan**

- [x] As a site user (anon), I can submit a search, and on the results screen, I can select one or more items on each of the Content Type and Group search facets to effectively narrow the search results.


lgbridges:
> 
> 
> @lgbridges search works great. Only UX issue that I'm having is that when I check a box, it disappears because 'show all' gets added and the active checkbox gets hidden under view more.
> 
> I would set 'sort by active state' in the facet sorting option to be sorted in ascending order instead of descending. That way, it's easier to disable a checkbox if user activates it.

Done. It's weird that ascending isn't the default for active state.....

jmilden-git:
I'll take QA.

jmilden-git (approved):
QA passes.

gatoona (approved):

PR #144, opened 2021-08-12
WCOR-1145: Simplify group entity reference fields.

gatoona:
### ODE 125

**Use Case** 

As an editor/admin, I would like to be able to reference content to a parent group more easily.

**Implementation Details**

- For the group content type, split parent group field into department, division, and program selection.
- For the displayed in field, display group path in addition to title and add searching via group path.

**Dependencies**

- Entity reference

**Test Plan**

- [x] Create a new news or page content type and add in web groups for the 'display in' field by searching via url paths.
- [x] Create a new group and select a department, division, and program for it.


jmilden-git:
Sample from QA test:
image
Would be nice if "Displayed In" appeared in the side column.  Also, breadcrumb should probably have 3rd level to ITS.

jmilden-git (changes requested):
Couldn't complete QA.  Last step needs clarification.  After creating a new Department group, I don't see any way to associate another sub group.

gatoona:
> Couldn't complete QA. Last step needs clarification. After creating a new Department group, I don't see any way to associate another sub group.

@jmilden-git There's a "Parent group (department)" field that will guide through the rest once it's filled in.

joshuami:
> Also, breadcrumb should probably have 3rd level to ITS.

Breadcrumbs are currently based on the path alias. We'll need to look at menu_breadcrumb and add all groups to a site menu if we want to override the breadcrumb when the path alias does not contain all the parents. I think @lgbridges was going to look into that as part of his story.

lgbridges:
> 
> 
> > Also, breadcrumb should probably have 3rd level to ITS.
> 
> Breadcrumbs are currently based on the path alias. We'll need to look at menu_breadcrumb and add all groups to a site menu if we want to override the breadcrumb when the path alias does not contain all the parents. I think @lgbridges was going to look into that as part of his story.

Right, will add Menu Breadcrumb as part of WCOR-1159. I would bypass all concerns with breadcrumbs in QA for the time being.

lgbridges:
@gatoona -- I understand the use case more fully now that we've talked. The goal here seems to be to allow us to embed all layers of hierarchy as separate parent department, parent division, parent program fields on a given group. This would presumably allow us to construct views or other queries in a "roll-up" fashion a bit more easily without having to traverse multiple levels of hierarchy (for example, a view that would show all programs for a given department).

joshuami:
@gatoona, great work pulling in the ideas we talked through. When I manually tested this one, it felt a little clunky from an editor's perspective. I've done a quick mock of an alternative (while waiting for Portland builds to complete 😄 ) that swaps the logic around a little bit: https://github.com/joshuami/COUNTY-DRUPAL/tree/feature/WCOR-1145. I also noticed a couple of interesting side effects from the change that we can talk through as well as an alternative to my alternative 😆 .

I haven't turned this into a PR just yet as I don't want to tie up a CD, but I'm happy to walk us through it after standup in the morning.

lgbridges (approved):

joshuami:
Shoot. We should have closed this one. It is incompatible with the PR I'm creating for the alternative approach. I'm glad I saw you approved the PR or I wouldn't have been reminded to close it. 😄

gatoona:
@joshuami thinking we might just ditch the conditional logic given that groups will only be edited by webteam. I might just reopen this PR as this included the new divided group fields and the ability to search for groups by path alias.

joshuami:
@gatoona , I have a branch that is just about ready for a PR, which is why I reassigned the story to me for completion. One thing that needs to be worked out is the conflict of removing `field_parent` which is a dependency for some of @lgbridges work to create a group contact page. If we could leave this on my plate for now, I can at least draw up an option for resolving that conflict so that we don't delete the field and result in breaking (deleting really) his views.

gatoona:
@joshuami I'm well aware of the field dependencies on Lee's story. I would figure the solution is as simple as just creating the field_department field and replacing that with field_group for the views Lee created. I have a few content migration pieces that's currently on hold until we get the group field separated.

joshuami:
There is some value to the conditional fields—even for the DigX team—as it would prevent the team from accidentally creating a group with improper hierarchy accidentally. The approach in my branch also groups those fields in to an administrative field group in the sidebar that really cleans up the form. I'd hate to lose that work or have you recreate it.

As for the views, `field_parent` that Lee's view is using. Just adding `field_department` instead doesn't quite address the issue. The usage of the relationships between a department and its divisions; a division and its programs; and a program and its subprograms will likely all need to be a backreferenced list view that could be displayed as a sidebar menu or a link to the page list view of the children. 

If you wanted to take https://github.com/joshuami/COUNTY-DRUPAL/tree/feature/WCOR-1145 and riff off that work there, I'd be cool with it, but I really don't mind finishing up the spike and demoing it to the team before we create that PR. It might be worth having a quick conversation about having the hierarchy tracked in a menu for ease in generating a sitemap as well. Lot's of options that we should probably talk through before just merging. 😄

PR #146, opened 2021-08-13
WCOR-1064: Install Views Bulk Operations and Views Bulk Edit

jmilden-git:
### ODE 126

**Use Case** 

As an admin, I would like to bulk update fields for selected content.

**Implementation Details**

Install and configure VBO & Bulk Edit. Also install patch needed for moderation states.  Adjust default system views to enable VBO Actions for selected items.  Add to admin Content view and Group Content and Media Library & Group Media.

**Dependencies**

- Patch file

**Test Plan**

As an administrator, 

- [x] Open default Content views and attempt to apply a moderation state (Action) to multiple nodes. 

- [x] Go to Media Library and ensure the bulk operations allow for modifying fields on media.


jmilden-git:
@joshuami & @gatoona  Finally got the CD Env (126) built for this one.  If one of you can take QA and the other code review. Thanks

config/default/views.view.manage_group_content.yml
```diff
@@ -120,6 +121,76 @@ display:
...
+              action_id: 'moderation_state_change:node'
+              preconfiguration:
+                add_confirmation: true
+            7:
+              action_id: views_bulk_edit
+              preconfiguration:
+                add_confirmation: true
+                get_bundles_from_results: true
+            8:
+              action_id: views_bulk_operations_delete_entity
```
joshuami, line 189: We probably want to leave the delete action off of manage group content and the my content views. The editor and publisher roles don't have permission to do this action. I can't remember if the permissions are smart enough to leave off the action if the user can't use it.

joshuami (changes requested):
One small change that we should consider making on the manage group content view. 

It might also make sense to go ahead and add the bulk moderation action to the "my content" view as many editors will use that view instead of the group manage content view.

config/default/views.view.manage_group_content.yml
jmilden-git: @joshuami  Those sound like good changes.  I'll make those and update the PR.

lgbridges: @jmilden-git Eager to get this merged as it has utility for group content cleanup needed prior to my "flatten hierarchical paths" story.

jmilden-git: @lgbridges @joshuami This PR (Views Bulk Ops) has been updated.

joshuami: Hmm. Did you push the commit? The last one I'm seeing was a merge in of develop, but no view changes.

jmilden-git: @joshuami I thought it included my changes.  I used the commands to merge Dev from the wiki, then I used the update branch button in this PR.  It reran the pipeline, so I thought it worked.  Do you have time to give me another command or two?

joshuami: Did you edit the manage group content view on your IDE environment and run `drush cex` to get the changes? You should have one more commit that just deletes the delete entity action at a minimum. We can do a follow up to add bulk operations to the my content view if you like.

joshuami: Oh... and you'll know your commit contains what it needs if this comment thread shows up as "outdated"—your commit should delete the line I commented against. (Hope that makes sense 😄 )

jmilden-git: Yes, I did all your suggested changes, ran drush cex and committed the changes.

joshuami: I'm guessing you had an order of operations miss. The CD environment still has the delete option on manage group content:

2021-08-17 at 4 45 PM

You'll need to reimplement the changes on your IDE and export again.

jmilden-git: @joshuami I only removed delete per the advice, "We probably want to leave the delete action off of manage group content and the my content views."  Was that incorrect?  Also, I'm at a loss for how to get the last commit into the PR.

joshuami (approved):

lgbridges:
@jmilden-git QA got stuck on the last item. I masqueraded as Eddie Editor and viewed Content. All that was available were 3 news items. Not sure what that's about, seems unrelated to this story. I was able to select 2 news items and do the "Modify field values" action. I was able to choose a field and enter a new value. However, when I clicked "Apply", nothing happened. I tried several different kinds of fields, nothing worked. Also tried unchecking "create new revision", that didn't help. I will try to do some more troubleshooting as time allows.

lgbridges:
> 
> 
> @jmilden-git QA got stuck on the last item. I masqueraded as Eddie Editor and viewed Content. All that was available were 3 news items. Not sure what that's about, seems unrelated to this story. I was able to select 2 news items and do the "Modify field values" action. I was able to choose a field and enter a new value. However, when I clicked "Apply", nothing happened. I tried several different kinds of fields, nothing worked. Also tried unchecking "create new revision", that didn't help. I will try to do some more troubleshooting as time allows.

Additionally, I noticed that the Editor and Publisher permissions on Views Bulk Edit weren't enabled. I did enable those in the ODE (note: this config needs to be applied and committed to the PR), and cleared cache, but it didn't enable me to apply multiple field bulk edit changes.

jmilden-git:
@lgbridges I'll have to update the PR with the permissions.  I'll let you know when it's ready for recheck of that last test.

lgbridges:
> 
> 
> > @jmilden-git QA got stuck on the last item. I masqueraded as Eddie Editor and viewed Content. All that was available were 3 news items. Not sure what that's about, seems unrelated to this story. I was able to select 2 news items and do the "Modify field values" action. I was able to choose a field and enter a new value. However, when I clicked "Apply", nothing happened. I tried several different kinds of fields, nothing worked. Also tried unchecking "create new revision", that didn't help. I will try to do some more troubleshooting as time allows.
> 
> Additionally, I noticed that the Editor and Publisher permissions on Views Bulk Edit weren't enabled. I did enable those in the ODE (note: this config needs to be applied and committed to the PR), and cleared cache, but it didn't enable me to apply multiple field bulk edit changes.

@jmilden-git, @joshuami, and Team -- I think I had a moment where I wasn't doing enough critical thinking about our intentions, and it should've stopped me from making the previous comments. The use case states that "As an admin, I would like to bulk update fields for selected content." I'm actually thinking we don't necessarily need to put Bulk Ops and Bulk Edit in the hands of editors. Does anyone disagree? Is there real utility in it, and if so, does it outweigh the risk of problematic behavior?

jmilden-git:
@lgbridges Good questions.  This entire story has been driven by Josh's ideas and suggestions.  That's why I added the editor/publisher permissions (which are currently not working).  Let's have a call if you want to discuss further.

jmilden-git:
@lgbridges Finally got it pushed up, pipelines currently running.  Removed final QA step.

lgbridges (approved):

PR #151, opened 2021-08-24
WCOR-1165: BLT custom command support for Lando.

gatoona:
### ODE XX

**Use/Fix Case** 

As a developer using Lando, I would like to be able to run custom blt commands.

**Implementation Details**

- Setup needed directories after build process via root.
- Change environment variables to support Lando.
- Specify composer version to match with IDE (2.1.5)

**Dependencies**

- Lando
- Docker

**Test Plan**

- [ ] Custom BLT commands ran via Lando `lando blt custom` complete successfully in Lando environments.


.lando.yml
```diff
@@ -7,11 +7,24 @@ config:
...
   blt:
     service: appserver
-    cmd: /app/vendor/bin/blt
\ No newline at end of file
+    cmd: /app/vendor/bin/blt
+services:
+  appserver:
+    overrides:
+      environment:
+        DRUSH_OPTIONS_URI: "https://county-drupal.lndo.site"
```
gatoona, line 22: @joshuami looks like there's an option in Lando to set a uri so that `drush uli` doesn't show 'default' as the base url.

PR #159, opened 2021-08-26
WCOR-1145: Refactor group views and menu blocks also form improvements.

joshuami:
### ODE 140

**Use/Fix Case** 

As an editor (or even as a community member), I want to be able to find a list of divisions for each department and programs for each division. Additionally, as a site administrator, I want to be able to set the parent and type of a group to have it appear in the correct places.

**Implementation Details**

This PR does a few things. First, it cleans up the group form display to use a details element for administrative fields and moves social and contact information into the sidebar.

Next, the Groups view has been updated to use contextual arguments for division and program views as well as for the contact page—though we might not need the contact page due to the next change.

The primary service location is now output as an "embedded" display mode and as is the building within. This allows for the output of all the necessary fields for address and the EVA links for map and transit.

**Dependencies**

- None

**Test Plan**

The data is not clean enough to look correct, but fixing the parent structure for a few programs to have their parent as a division instead of the department shows the proper nesting.

After this is deployed to dev, we should clean up unneeded "administration" groups and map all subprograms and programs to their appropriate parents so that the entire structure shows correct relationships.

We should also update the spec tool with the new view structure and additional field definition.

- [x] Update WIC (group) to belong to the Public Health (group). You should then be able to traverse from the Departments list to see the Public Health Division under Health and then to the WIC program from the Public Health Division homepage.
- [x] Verify that the service location for WIC (group) shows the address and get directions links.

I can also demo this during architecure and planning tomorrow to talk through clean up steps for the data. 😄


config/default/simplesamlphp_auth.settings.yml
```diff
@@ -1,10 +1,10 @@
 langcode: en
-activate: true
+activate: false
```
lgbridges, line 2: @joshuami -- Quick check-in on this item, on face value this looks like a global deactivation for SimpleSAML, do you have any additional context to share?

joshuami (commented):
There you go. Files restored. Should pass now. Not sure how I ended up with those changes on my local. It seems to execute a `config-export` without producing them again. 🤔

lgbridges (approved):

PR #161, opened 2021-08-30
WCOR-1126: FAQ content type.

joshuami:
### ODE 141

**Use/Fix Case** 

As a community member, I want to be able to find answers to common questions either through search or when I find a service and have additional questions.

As an editor, I want to be able to create question answer pairs that appear on a page and in search.

**Implementation Details**

- Create FAQ content type from spec

- Show FAQs on the page content type

- Show FAQs in search

**Dependencies**

- Inline Entity Form

**Test Plan**

- [x] Create a page and embed several new FAQs into the FAQ field.
- [x] Create a second page and embed an existing FAQ into the FAQ field. Include a keyword not used in the page in the keywords field.
- [x] Test search using the keyword.


config/default/simplesamlphp_auth.settings.yml
```diff
@@ -1,10 +1,10 @@
 langcode: en
-activate: true
+activate: false
```
gatoona, line 2: @joshuami looks like this is overriding the default settings for simplesamlphp_auth.

joshuami: So it was a slightly different change between the two branches. On the first PR it picked up this file plus another config file related to SAML on the local config split. Darned if I can figure out how I did the same thing with different results or what order of operations I missed to cause this one. At least it is an easy fix.

gatoona (approved):

lgbridges (approved):

PR #163, opened 2021-09-01
WCOR-1155: On a page add service address from field location.

jmilden-git:
### ODE 143

**Use/Fix Case** 

As an editor I need to associate the correct service address or location to my page.  As a visitor I want to see the service location on the content page. 

**Implementation Details**

Create display modes for the referenced building and service_location entities that you display as a field on your group. 
Here is an example of a service location displaying with a mode of "related": https://github.com/eGovPDX/portlandor/blob/master/web/themes/custom/cloudy/templates/content/service-location/node--service-location--related.html.twig.Note that Portland.gov uses content types called service_location and location. So where you see the word location, you would want to change that to the machine name "building". 

You can also create a building template so that either a service location or a building could be used by field_location: https://github.com/eGovPDX/portlandor/blob/master/web/themes/custom/cloudy/templates/content/location/node--location--related.html.twig

For the display of the page (or group or news) using a service_location or building, you just need to output the field and display it as a rendered entity instead of a label.

For now, the placement of the field should be in the main body. We can do some work with layout builder or other approaches in the future that would place that information into a side region. We need a bit more work with IF to work out the strategy for layout before we do that kind of placement just yet.

**Dependencies**

-Service Location

**Test Plan**

- [ ] Create a new Page with test content
- [ ] Associate a service location in the field Primary Service Location.
- [ ] Check to see the address information is displayed on the page.
- [ ] Create a new Group with test content
- [ ] Associate a service location in the field Primary Service Location.
- [ ] Check to see the address information is displayed on the page.


config/default/core.entity_form_display.node.page.default.yml
```diff
@@ -6,6 +6,7 @@ dependencies:
     - field.field.node.page.field_body
     - field.field.node.page.field_group
     - field.field.node.page.field_menu_parent
+    - field.field.node.page.field_primary_service_location
```
lgbridges, line 9: @jmilden-git -- I've been gaining awareness of places where we should be keeping relational fields more open-ended. For a page, we would want the ability to reference multiple locations (and possibly either a building or service location). I believe we need to throw out field_primary_service_location in favor of a plain Location (field_location) field that can allow multiple values and can be populated with Building or Service Location items.

jmilden-git: @lgbridges Should the field_location be an entity reference, so point to a new content type like Service Location, but called Field Location?

jmilden-git:
Try with new PR for same feature.

PR #165, opened 2021-09-02
WCOR-1173: Simplify group reference search.

gatoona:
### ODE 144

**Use/Fix Case** 

With the large amount of groups and duplicate names, I would like to be able narrow down the group(s) I want to select in 'displayed in' and 'parent group' fields by searching for paths in addition to titles.

**Implementation Details**

- Update group reference view to include paths and also allow for searching via paths.
- Implement it for FAQ displayed in.

**Test Plan**

- [x] When creating a new group, ensure that parent group results can be narrowed down using a path.
- [x] When creating a new FAQ, ensure that displayed in results can be narrowed down using a path.


lgbridges:
@gatoona -- I agree that it would be helpful to show the nested hierarchy in the autocomplete for parent group selection (and make it searchable) but the changes I'm planning for the contents of Group Path field will end up diluting this usefulness of this function. That may be a non-issue since group setup is a site admin function.

The plan is to simplify paths--for example, _/lut/planning-and-development-services/long-range-planning/transportation-planning_ becomes _/transportation-planning_ or _/lut-transportation-planning_. This should make page paths more accessible, more typeable when necessary, more print-friendly, etc.

I just want to be transparent that I'm definitely wanting to pursue simplification of group path, and that it will likely impact the effectiveness of this feature. However, if we want to treat this as a tool for our initial setup of group relationships as site admins before we start cleaning up group path, that would make sense. We could revise the feature later, maybe pivot it toward some kind of recursive query to grab nested parent group titles and use that to represent the hierarchy in the view instead of group path.

Just wanted to voice that concern. Let me know if you want to discuss further.

gatoona:
@lgbridges all okay and I agree with simplifying the URL path. I don't think it will have any impact related to this feature even after we simplify the URL because there's still going to be a need to be able to easily select a group from the autocomplete. This just adds onto the ability to search for a group by allowing paths to also be searchable. If an author wants to have a page displayed in a certain group, they can either search via title or current path.

lgbridges (approved):

PR #166, opened 2021-09-03
WCOR-1168: Implement emergency closures for locations.

lgbridges:
### ODE 145

**Use case**

As a Content Admin (new role), I can add a Closure Event and associate a Location, and while adding/editing a News item, I can add a related Closure Event, and this will trigger display of the closure date range on the event-associated location (and subordinates).

As a site visitor, while viewing office hours for a service location, I can see a Closure announcement line just above the office hours, linked to the related Event item.

As a site admin, I can delete a Closure Event.

**Implementation plan**

Implement the suggestions made in the "Potential approach" notes of SFSTRY0001128.

**Test plan**

- [x] Add user Catherine Content-Admin. Ensure content admin role is assigned. Use "no-reply4@co.washington.or.us" as email address. Assign any secure password (no need to record it).
- [x] Masquerade as Eddie Editor. Open the Add Content page and notice that Event is not an option. Unmasquerade.
- [x] Masquerade as Catherine Content-Admin. Add an Event called "PSA Closure". Assign a date range that includes the current time. Choose Public Services Annex as the Location. Choose Published status and save. Add a News item called "Public Services Annex closed to the public". Choose any News type. Displayed In can be set to Public Health. Be sure to set Related field to "PSA Closure" event. Choose Published status and save. Unmasquerade as needed.
- [x] As a site visitor (anon), find and view the Hillsboro WIC service location and see the "PSA Closure" notification (including date range) above the hours. Click it to view the related Event item.


gatoona:
@lgbridges all test tasks are completed. One item that I had a question on was the news content type that was created. So the closure event is linked from the news item, but the news item is not linked anywhere within the event. Is this something that will be added later or was missed in this story?

lgbridges:
> 
> 
> @lgbridges all test tasks are completed. One item that I had a question on was the news content type that was created. So the closure event is linked from the news item, but the news item is not linked anywhere within the event. Is this something that will be added later or was missed in this story?

@gatoona -- I didn't really consider it for this story since nobody will be paying much attention to the events themselves initially since their only function is closures. Actually, what I would prefer to happen would be that the clickable closure link would lead the user to the news item, not the event. This would take a more complex view and I just ran out of time. I will add a story for next sprint to handle that.

gatoona:
@lgbridges gotcha. If you want the closure links/urls to redirect to the news item, that's actually pretty easy to do using the rabbit hole module.

gatoona (approved):

gatoona (approved):

PR #185, opened 2021-09-11
WCOR-1155: Add Location field to page content type.

jmilden-git:
### ODE 160

**Use/Fix Case** 

As an editor I need to associate the correct service address or location to my page.  As a visitor I want to see the service location on the content page. 

**Implementation Details**

For the display of the page using a field_location or building, you just need to output the field and display it as a rendered entity instead of a label.

For now, the placement of the field should be in the main body. We can do some work with layout builder or other approaches in the future that would place that information into a side region. We need a bit more work with IF to work out the strategy for layout before we do that kind of placement just yet.

**Dependencies**

-Building and Field Location

**Test Plan**

- [x] Create a new Page with test content
- [x] Associate a service location in the field Location, save and view the page.
- [x] Check to see the address information is displayed on the page.


gatoona (approved):
Created test page at /hhs/test.
Address is displayed.

PR #192, opened 2021-09-17
WCOR-1174: Initial Solr indexing configuration settings.

lgbridges:
Initial spike / research story to understand and document Solr config settings and starter values inspired by Josh's demo of Portland.gov Solr indexing.

Initial Solr index settings to match the docs have been created and submitted in this PR. There is no test plan, as a separate story will be created after content migration is well underway to rigorously test search and adjust settings as needed.


gatoona (approved):
CR passes.

PR #195, opened 2021-09-27
WCOR-1192: Purge old aliases when content is saved.

lgbridges:
### Use case

As an author, when a change occurs to a field that is used by Pathauto patterns to generate URL aliases, the node alias will be updated and all prior aliases for the node will be dropped.

### Implementation details

Enable the Pathauto setting "Create new alias. Delete old."  Also, remove "node:source:" in patterns and replace with "node:".

### Test plan

As site admin --

- [x] Find (or create) a page with "Displayed in" populated. Make sure the group populated there is published (eg. WIC).
- [x] Find the group populated in "Displayed in" for the previous page and change its group path.
- [x] Masquerade as Eddie Editor. Make any change to the previously mentioned page. Set publication status to either Draft or In Review.
- [x] Unmasquerade. Move page to Published status.
- [x] Check the list of aliases (/admin/config/search/path) to confirm that the old alias is deleted and the new alias (based on changed group path) now exists.


sharstine:
@lgbridges - I was able to follow the steps in the test plan, but the path of the page I updated did not change as I expected.

Here's what I did:

- Find (or create) a page with "Displayed in" populated. Make sure the group populated there is published (eg. WIC).
  - /wic/breastfeeding-support

- Find the group populated in "Displayed in" for the previous page and change its group path.
  - Changed URL alias on WIC group (/wic) to /qatest

- Masquerade as Eddie Editor. Make any change to the previously mentioned page. Set publication status to either Draft or In Review.
  - Completed masquerade as Eddie Editor
  - On /wic/breastfeeding-support, I added "QA test change" to the first paragraph under the image
  - I set the publication status to "In Review"

- Unmasquerade. Move page to Published status.
  - Completed unmasquerade
  - Set /wic/breastfeeding-support to "Published"

When I go to /admin/config/search/path, I see that /wic no longer exists (the alias is now /qatest). But the page I edited and published still has the alias /wic/breastfeeding-support. If I understand the use case correctly, I would have expected the alias of the page to change to /qatest/breastfeeding-support. Please let me know if I misunderstand the use case or missed a step in QA.

As a final step, I cleared all caches and ran cron but observed no change to the URL aliases afterward.

lgbridges:
> 
> 
> @lgbridges - I was able to follow the steps in the test plan, but the path of the page I updated did not change as I expected.
> 
> Here's what I did:

Somehow, it looks like the Pathauto "update action" setting I had changed didn't get committed. I will push that out and check the test plan again.

lgbridges:
Updating linkit module here to circumvent Pipelines security flag. Would typically handle through hotfix PR but we are out of CD environments and I didn't want to destroy current QA work.

lgbridges:
I have verified the test plan in my local after the latest changes. Ready for QA.

sharstine:
@lgbridges - Unfortunately, I'm getting identical results as yesterday on ODE 171. I tested with the same page (/wic/breastfeeding-support) and WIC group (/wic, which I re-aliased /qatest2). Ultimately, the page I edited and published still has the alias /wic/breastfeeding-support. I cleared all caches and ran cron after testing the first time, but observed no change in the URL alias afterward.

Today I also tested with a different set of data and had the same outcome:

Page: /public-health/test-page-qa
Group: /public-health (I changed the alias to /ph)

lgbridges:
> 
> 
> @lgbridges - Unfortunately, I'm getting results identical to yesterday on ODE 171. I tested with the same page (/wic/breastfeeding-support) and WIC group (/wic, which I re-aliased /qatest2). Ultimately, the page I edited and published still has the alias /wic/breastfeeding-support. I cleared all caches and ran cron after testing the first time, but observed no change in the URL alias afterward.
> 
> Today I also tested with a different set of data and had the same outcome:
> 
> Page: /public-health/test-page-qa Group: /public-health (I changed the alias to /ph)

@sharstine -- I reviewed your work in ODE171 and I think you were changing the URL alias for the group, not the Group Path. Pathauto is deriving Level 1 paths in URL aliases from the group path field that we added, not the stock URL alias field. See screenshot below for the field you'd need to update on Step 2 of the Test Plan
Capture
.

sharstine:
@lgbridges - Thank you! I definitely edited the wrong field the first two times. I re-ran the test plan and it passes. QA is complete.

gatoona (approved):
CR passes

PR #238, opened 2021-10-28
WCOR-1206: Implement CKeditor changes & page element styles.

lgbridges:
### Use case

As a site admin, I want to streamline the CKeditor experience and build up styles for a couple standard page elements (namely blockquote, lists, and list items).

### Implementation plan

Update CKeditor toolbar config and add styles.

### Test plan

As Eddie Editor --

- [x] Create a test page. Note the lack of a horizontal line button in CKeditor.
- [x] Add an Image from Media Library. Enter source view and note the addition of a `<drupal-media>` tag.
- [x] While in source, add a valid `` tag that calls an image. Note that, upon leaving source view, the image is not presented in WYSIWYG view. Go back into source and see a `<p>&nbsp;</p>` placeholder where your `` tag was placed.
- [x] In WYSIWYG view, add an ordered list with at least 4 tiers (indentions) of items. Do the same for an unordered list. Note that the indentions are all uniform and even.
- [x] Add a blockquote and observe alternate styling emphasizing the text in the blockquote element.


sharstine:
@lgbridges - QA passes. Also confirmed that styling on the published page appears to match styling in CKEditor.

A few changes to consider/address perhaps in a follow-up story:

- The text in the blockquote doesn't wrap and gets cut off the page in mobile view.
- Ordered lists in CKEditor default to roman numerals. Is this the standard we're going for? If not, reconfigure so that standard numbers are the default.
- Top-level list items are currently not indented. Should they be?
- When clicking the mouse at the end of a list item (ordered or unordered) in Chrome, sometimes the cursor is placed at the beginning (rather than end) of the list item. I didn't test in other browsers. You can use the arrows on the keyboard to move the cursor to the end of the list item, but we should improve the experience for authors if possible.

jlwcor (approved):

PR #239, opened 2021-10-29
WCOR-1216: Menu toolbar for admin views.

lgbridges:
### Use case

As a site admin, I want to easily access all admin-only views.

### Implementation plan

Create Admin Views menu, configure Toolbar Menu to display it, and ensure correct permissions are set.

### Test plan

As a site admin --

- [x] Click the "Admin Views" menu in the toolbar. Click on "Migration status" item and see the Content Migration Status view.


sharstine:
I'm confused about why the bot requested me as a reviewer and then removed me.

Assigning myself as QA person and starting review now.

washco-or:
> I'm confused about why the bot requested me as a reviewer and then removed me.
> 
> Assigning myself as QA person and starting review now.

sharstine:
QA passes.

As we get deeper into content migration, I imagine we'll want to add additional filters or other enhancements to the Content Migration Status page. Some things to consider for a later time:

- The Content Migration Status page only displays pages where the currently logged in administrator is named in the "Authored by" field. If someone else (e.g., Surya) creates our pages initially, we won't see them in the current view. We may need a filter for "Authored by." Also, maybe an option to see all pages regardless of "Authored by" value.

- The Content Migration Status page displays pages that don't have a Migration Status set. Should we consider making the Migration Status field required?

- I forgot how/where to do this, but Josh showed us a way to search nodes by string. If we consistently insert "[addlink]" into our pages as placeholder text when we need to come back at a later time to add a link, for example, we can easily find those pages. It may be helpful to incorporate that search function into the Content Migration Status page.

jlwcor (approved):

PR #245, opened 2021-11-02
WCOR-1205: News search.

lgbridges:
### Use case

As a site visitor, I want to search the full text of news items. I also want to be able to narrow the full list or the search results by year, month, and group.

### Implementation plan

Implement search page in Views. Configure facets and facet block display for year, month, and group. Appearance of Month facet should be dependent on a value selected in Year facet.

### Test plan

As a site visitor (anon) --

- [x] Go to /news and enter a search in the Search filter field. Observe the new results set.
- [x] In the Year facet, select a year and observe the Month facet appear upon page reload.
- [x] In the Month facet, select a month, and in the Group facet, select a group. Note the filtering of results.


jmilden-git:
QA passes as written.  However, this strange "Show all" option appears after selecting a facet.  Yet, it only inludes the same items as showing in the filter.  This seems redundant / confusing, since it doesn't really show all.  Can you remove the Show all filter option?
image

lgbridges:
> QA passes as written. However, this strange "Show all" option appears after selecting a facet. Yet, it only inludes the same items as showing in the filter. This seems redundant / confusing, since it doesn't really show all. Can you remove the Show all filter option? 

This is how the facet configuration handles a "reset" function for the facet. The default text is "Show all" but could be changed to "Reset". However, deploying the facet to use checkboxes (the preferred method) means that the Reset just appears as a checkbox alongside any selected options. I would prefer Reset to be a separate link, perhaps at the bottom of the facet block, etc.

We could leave the reset function in (renaming it if we wish), or we can remove it if it seems confusing. I've also found through my own testing both here and on the Portland project that the Reset function also occasionally seems a bit glitchy -- sometimes when used, after page reload some facets will be missing, or missing options.

I'm fine with disabling the internal facet "Reset" for now. I'm going to add to Issues queue a request to add a facets *global* Reset as a custom link in the region that will effectively reset all facets by reloading the page, just including the current search query (the `search` URL query param).

joshuami:
I agree with the wonkiness of the facet reset link. I've always though it would be better implemented like the soft limit show more/less links. 

One thing that can cause facets to be a bit strange is when you have dependent facets, such as when you have the year conditionally show months for that year.

An option that might feel a little cleaner would be to turn on the reset button for the exposed form (your search box) and hide the facet resets. The exposed form reset clears all URL parameters, which functionally is what the show all link is doing, but only for a single facet.

lgbridges:
Removed individual facet resets.

washco-or (approved):

PR #247, opened 2021-11-02
WCOR-0: Enhance Content Migration Status admin view.

lgbridges:
## ODE 213

Adjustments based on feedback in #239. 

### Implementation Details

- Add exposed Authored By filter to allow viewing content authored by any site admin.
- Require value in exposed Migration Status filter.
- Add exposed Body "Contains" filter to allow search for custom/user-defined tokens such as [addlink] that act as reminders for follow-up work.

### Test Plan

As site admin --

- [x] Open view-page at /admin/content/migration and try viewing content entered by other admins. Check functionality of Migration Status filter.
- [x] Open a page and add a custom/user-defined token (e.g. [addlink], [addimage]). Return to the view and search for the item using the exposed filter.


sharstine:
@lgbridges - QA passes. This dashboard will come in handy!

Some additional future enhancement ideas below. I think we can get by without them for now, but they might be worth considering later on -- depending on how we use the dashboard.

* Add "Any" or "All" option to the Migration Status filter. I may want to filter by [addlink] across all my pages, regardless of migration status. Also, since Migration Status is currently not a required field, the Migration Status filter could leave some pages out of view.

* Would be nice to have a "View All" option near the pager to see all filtered pages in a single view. This would allow us to Ctrl+F to find a particular page title.

* The pager is a bit misleading to me. Joe has ~26 pages with the status "In migration." At the bottom of the first page of results, the pager shows "1." Aside from a forward arrow, there's no indication that there's a page 2 and/or how many total pages of results there are. I think the DigEx team can live with this, but if we roll out pagers in author dashboards it would be nice to have a more intuitive pager that clearly suggests there are additional pages to click through. The pagers on Google and Horizons are examples of how it could be improved.

joshuami:
@lgbridges, having seen @sharstine 's comments, any/all, and a "full pager" are doable with a couple of checkboxes on the view. I might be able to show you Monday if you like. I would avoid the view all option for the pager. You'll likely have thousands of pages at some point and that would be a very slow query. 😄 

Have you considered just adding the migration status filter to /admin/content instead of creating a new view? It would forgo the need for a new editor toolbar menu and make is part of overall management. 🤔

lgbridges:
> @lgbridges - QA passes. This dashboard will come in handy!
> 
> Some additional future enhancement ideas below. I think we can get by without them for now, but they might be worth considering later on -- depending on how we use the dashboard.
> 
> * Add "Any" or "All" option to the Migration Status filter. I may want to filter by [addlink] across all my pages, regardless of migration status. Also, since Migration Status is currently not a required field, the Migration Status filter could leave some pages out of view.
> * Would be nice to have a "View All" option near the pager to see all filtered pages in a single view. This would allow us to Ctrl+F to find a particular page title.
> * The pager is a bit misleading to me. Joe has ~26 pages with the status "In migration." At the bottom of the first page of results, the pager shows "1." Aside from a forward arrow, there's no indication that there's a page 2 and/or how many total pages of results there are. I think the DigEx team can live with this, but if we roll out pagers in author dashboards it would be nice to have a more intuitive pager that clearly suggests there are additional pages to click through. The pagers on Google and Horizons are examples of how it could be improved.

These requested changes have been made.

lgbridges:
> @lgbridges, having seen @sharstine 's comments, any/all, and a "full pager" are doable with a couple of checkboxes on the view. I might be able to show you Monday if you like. I would avoid the view all option for the pager. You'll likely have thousands of pages at some point and that would be a very slow query. 😄
> 
> Have you considered just adding the migration status filter to /admin/content instead of creating a new view? It would forgo the need for a new editor toolbar menu and make is part of overall management. 🤔

On the last item -- this sounds more streamlined, but for some reason I'm hesitating. The migration view is a temporary thing that will go away upon Phase 1 launch.  I also don't want to load up the exposed filters too much -- being different use cases, I think each view might be more functional on their own.

Maybe after we work with it a bit we'll change our mind about how we want admin views to work.

washco-or (approved):

PR #258, opened 2021-11-10
bugfix/WCOR-0: Disable EVA phone formatting.

jlwcor:
### ODE 219

**Use/Fix Case** 

Replace EVA for phone formatting rendering. Use theme twigs instead.

**Implementation Details**

- Disable EVA for phone formatting.
- Implement custom twig hook to format phone numbers.
- Use twig_tweak to render fields and referenced entity nodes.
- Disable individual contact node pages.

**Dependencies**

- drupal/twig_tweak

**Test Plan**

- [x] Phone number/extension information displays correctly in group content types.
- [x] Phone number/extension information displays correctly in page content types.


sharstine:
@jlwcor - The Pipelines job failed. I'll begin QA when the CD site is built.

sharstine:
@jlwcor - QA passes.

Ideas for future enhancement:

- **Phone number field:** Add validation to check for fewer than 7 digits being entered. I was able to save the field by entering just "503846." This may not need to be addressed at all. I think it's unlikely that somebody would _not_ enter a complete phone number, and phone numbers don't change often.

- **Extension field:** Disallow spaces, special characters, and alphabetic characters from being entered. When I enter a space and no other text into the field, an "x" still shows up on the published page.

lgbridges (approved):

PR #259, opened 2021-11-11
WCOR-0: Page presentation and pathauto bugfixes.

lgbridges:
### Use case

As a site visitor, pages should display with a summary in enlarged type and with adequate whitespace around it, and extraneous tracking fields such as Displayed In and Migration status should not appear.

As a content editor, I need URL aliases to not include slashes that appear in page's short title, creating the false appearance of additional hierarchy in paths.

### Implementation details

- Adjust page content styles.
- Adjust Page default display mode config.
- Adjust Pathauto settings.
- Update content_migration_status admin view to avoid setting default value for exposed filter 'Migration status', and make the filter not required.

### Test plan

As Eddie Editor or site admin --

- [ ] Update or create a page. Put a slash somewhere in the Short Title. Ensure Summary is filled in with 1-2 sentences.
- [ ] View the page and observe larger type for Summary and more whitespace. Confirm there are no visible tracking fields such as Displayed In, Type, Service Type, or Migration Status. Also, confirm the page URL contains no additional slashes past the Group Path.

As site admin --

- [ ] Viewing /admin/content/migration, there should be no default value set for exposed filter Migration Status, and the filter should not be required.


mitziklambert:
QA Completed

washco-or (approved):

PR #265, opened 2021-11-17
WCOR-0: Support for behat testing in cloud IDE.

jlwcor:
### ODE XX

**Use/Fix Case** 

As a developer, I would like to run behat tests in cloud IDE.

**Test Plan**

- [ ] Copy /home/ide/project/blt/example.local.blt.yml as /home/ide/project/blt/local.blt.yml
- [ ] Uncomment project configuration and set as:

```
project:
 local:
   protocol: http
   hostname: localhost
```

- [ ] All behat tests pass successfully after running `blt tests:behat:run`


jlwcor:
Need to pull in develop database via blt prior to starting behat testing.

washco-or (dismissed):

lgbridges:
@jlwcor -- We're seeing increasing volume in the PR queue lately and this has been here for a few sprints. We should talk more generally about priority for the Behat testing, but for now would you be willing to stash this?

PR #269, opened 2021-11-19
WCOR-1217: Convert social media EVAs to Twig rendering.

mitziklambert:
### ODE 225

**Use/Fix Case** 

Depts/divisions need to display social media information on group pages.

**Implementation Details**

On group content type, created twig field templates for each social media field instead of displaying them via EVAs (entity views attachments). 

**Dependencies**

The social media fields are conditional and only show up if there is social media info present within the group.

**Test Plan**

Navigate to the Solid Waste and Recycling (SWR) group page. (This group has both a Facebook page and YouTube channel, which is why we are using it.) Edit the group and enter the following into the right side under "Social Media" tab:
Facebook: washingtoncountyrecycles/
YouTube: /channel/UCEy5ulJ1a9K-WfmmUlsBvSg

After saving the above information into the group page, verify that links are displaying correctly on the Group page. The links should read 'Facebook' and 'YouTube' near the bottom of the page, but the links should take a visitor directly to the SWR social media accounts.


lgbridges:
CR passes; bugfix to follow to remove social_media_formatter view.

sharstine:
@mitziklambert - QA passes.

Some suggestions for clean-up:

* The test plan in this PR and help text in the YouTube field advise to add a leading "/" before "channel/UCEy5ulJ1a9K-WfmmUlsBvSg." This doesn't create a problem technically, but it results in the hyperlink having an extra "/" character after the YouTube domain. For example: http://youtube.com//channel/UCEy5ulJ1a9K-WfmmUlsBvSg. I'd suggest updating the help text to remove the leading "/" in the example.

* When Facebook, YouTube, Twitter, and a phone number are populated, the phone number is currently displayed between YouTube and Twitter on the published group page. I'd suggest displaying all the social media fields together and moving the phone number up -- probably under email address.

lgbridges (approved):
@mitziklambert -- We can go ahead and merge this, although I'd recommend a follow-up bugfix PR to address Scott's QA feedback. Let me know if you have any questions.

mitziklambert:
@sharstine, I will start a bug fix for this per your recommendations. Jung was working on the phone number field at the same time I was working on the social media fields so there is obviously something out of order here. I will also work on removing the beginning / from the help text on the social media fields in the group content type.

PR #271, opened 2021-11-24
WCOR-1222: Group documents list / search view.

lgbridges:
### Use case

As a site visitor, I would like the ability to view long lists of document links (such as those on Finance pages) in a more functional way, including sorts, grouping, and faceted filtering.

### Implementation plan

Create a view-page for querying Solr and presenting documents with typical Solr capable components. The page takes first-level path as a contextual filter to filter by field_group, and allows for faceting by terms from Tags taxonomy and by field_fiscal_year value. The table will be grouped by the Tags terms, and a second column displays the Fiscal Year value (if any).

### Test plan

- [x] Prepare a set of content to match an example such as [Finance Reports](https://www.co.washington.or.us/Support_Services/Finance/CountyFinancials/financial-reports.cfm) from Finance group. Matching the example page, begin by adding the term "Financial Reports" to the Tags taxonomy, and then add the term "Component Units" as a child term to the Financial Reports term.
- [x] Create 4 test documents using any dummy PDF file. All documents should have Displayed In value of "Finance".
   - One document should be named Comprehensive Annual Financial Report, have Financial Reports as a Tags value, and Fiscal Year 2020.
   - The next document should be the same as above, but with Fiscal Year 2019.
   - The next document should be named Enhanced Sheriff Patrol District, have Component Units as a Tags value, and Fiscal Year 2020. 
   - The next document should be the same as above, but with Fiscal Year 2019.
- [x] Visit path /finance/documents to see a list view of the documents. Note that when you filter by facet, a new URL is generated that can be used to present a specifically filtered list of documents (used to create a Resource for a persistent view, etc).


sharstine:
@lgbridges - QA passes. Looks great!

Some modifications to consider:

* Non-selected filter checkboxes disappear when I select a tag or year. For example, when I select the "Component Units" filter, the "Financial Reports" filter disappears. If there were 3 tags and I wanted to filter by 2 of those tags, there's currently no way to see all the matching documents in a single view.

* On mobile, the filters are currently at the bottom of the page, which could be hard to find following a long list of documents. One option is to move the filters to the top of the body area, but doing so would push down the list of documents. An alternative would be a "Filters" button on mobile like the City of Portland is doing on mobile here: https://www.portland.gov/search?keys=policy&op=

* On mobile, the fiscal years hang off the right edge of the page a bit. I imagine this will be corrected when we apply site-wide table styling.

jlwcor:
@lgbridges quick question about the new FY field. Is this going to be set for all documents? For example, for my bulk PDF upload story, do I need to populate that field?

lgbridges:
> @lgbridges quick question about the new FY field. Is this going to be set for all documents? For example, for my bulk PDF upload story, do I need to populate that field?

@jlwcor -- No, Fiscal Year is an part of the unstructured data of CommonSpot and not something we can reasonable package for import. This is something that we & authors will have to manually populate.

jlwcor (approved):

Issue #281, opened 2021-12-08
Remove webform content type.


lgbridges:
Webform links will refer users to a webform on our on-prem Forms Server (including a Markup section at the beginning of form to resemble a page, if body content is desired). 

We should remove this content type until the effort to migrate forms begins after Phase 1.

PR #286, opened 2021-12-13
WCOR-0: Remove webform module and related placeholder content type.

lgbridges:
### Use/Fix Case

From issue #281: 

Due to the time budget to fully implement, test, and secure the feature, and given the availability of the on-prem Drupal Forms Server, DigEx will not implement webform within the Acquia instance of County Site in Phase 1 scope. We've discussed the mechanism for redirecting traffic for form use elsewhere.

Webform is therefore being removed from this Drupal instance for the time being. This is required to unblock existing PRs since webform needs a security update.

### Implementation Details

- Remove "webform" content type.
- Remove webform module.

### Test Plan

- [ ] Ensure site loads.
- [ ] Create any page and ensure it can be viewed.


jmilden-git:
@lgbridges  Why not just apply the security update?  Seems like a lot less work than removing and then reinstalling later.  We need web forms for new requests.

jlwcor:
I also agree with @jmilden-git that it would be better to keep the module installed and updated even though we are not utilizing it 100% right now.

PR #296, opened 2021-12-14
WCOR-1233: Update core to 9.3.0, handle multiple module updates.

lgbridges:
### Use/fix case

Begin testing against Drupal 9.3 and other updates. Webform module requires a security update blocking Pipelines.

### Implementation details

Please review the [9.3 update notes](https://www.drupal.org/project/drupal/releases/9.3.0). Core has applied a different sort to config export YAML files, which explains much of the diff going into the repo. 

We should also begin discussing an upgrade to PHP 8, as it is recommended now.

### Test plan

- [x] Load site.
- [x] Create a page and view it successfully.


lgbridges:
Search API and Search API Solr seeing issues upon update; needs research. Will pursue separately.

lgbridges:
Feeds update seeing issues upon update; needs research. Will pursue separately.

lgbridges:
Shield update seeing issues upon update; needs research. Will pursue separately.

jlwcor:
@lgbridges quick question. What were the steps you took to update the webform module libraries?

lgbridges:
> @lgbridges quick question. What were the steps you took to update the webform module libraries?

@jlwcor I'm pretty sure it was `composer update drupal/webform --with-all-dependencies; drush updb; drush cr; drush cex` and commit. At that point, Pipelines wasn't happy with it though and seemed to need updates for 4 specific ckeditor dependencies, so I handled those separately (see the commit that follows).

lgbridges:
> @lgbridges quick question. What were the steps you took to update the webform module libraries?

@jlwcor I noted the 4 ckeditor libraries that Pipelines was complaining about and updated them individually with `composer update --with-all-dependencies`, then pushed that out as a separate commit after the webform update.

lgbridges:
> > @lgbridges quick question. What were the steps you took to update the webform module libraries?
> 
> @jlwcor I noted the 4 ckeditor libraries that Pipelines was complaining about and updated them individually with `composer update --with-all-dependencies`, then pushed that out as a separate commit after the webform update.

I see now that the version doesn't seem updated according to the status report in the ODE.

composer.json
```diff
@@ -28,13 +28,14 @@
     "drupal/captcha": "^1.2",
     "drupal/components": "^2.4",
     "drupal/config_split": "^1.7",
+    "drupal/core": "^9.3",
```
jlwcor, line 31: This shouldn't be in here as we are already using drupal/core-recommended

composer.json
```diff
@@ -28,13 +28,14 @@
...
     "drupal/core-composer-scaffold": "^9",
-    "drupal/core-recommended": "9.2.10",
+    "drupal/core-recommended": "^9.3",
```
jlwcor, line 33: Please keep version for core-recommended fix otherwise during pipelines install, it will install a different version.

jlwcor (changes requested):

composer.json
lgbridges: @jlwcor So I followed the gospel instructions at https://www.drupal.org/docs/updating-drupal/updating-drupal-core-via-composer. For installs using core-recommended, they call for `composer update drupal/core "drupal/core-*" --with-all-dependencies`. But I see that this has added drupal/core to composer.json where that didn't exist before.

So you're saying this should've just been `composer update drupal/core-recommended --with-all-dependencies`?

composer.json
lgbridges: OK, so `9.3` instead of `^9.3`.

lgbridges: I will push this change.

sharstine:
@lgbridges - QA passes according to test plan. I created a test page (/hhs/test-page-qa-scott) with formatted text, document links, an image, and a video. The page rendered as expected as an admin and anonymous user. I tested several other miscellaneous things in Drupal and didn't encounter any issues.

lgbridges:
Superseded by #301.

PR #300, opened 2021-12-15
WCOR-1231: CKEditor Accordion Widget Tool

jlwcor:
### ODE 233

**Use/Fix Case** 

As a user, I would like to add accordion components into my page through CKEditor.

**Implementation Details**

- Develop and install custom ckeditor accordion module.

**Test Plan**

- [x] Using CKEditor, I am able to insert an accordion into the body using the accordion widget tool (button is a box with two arrows expanding inside it.)
- [x] Both collapsed and non-collapsed states can be toggled through the tool.


jlwcor:
@lgbridges @sharstine PR is ready for review.

lgbridges:
@jlwcor QA passes. This looks fine to me for MVP release.

I am noting that there is a difference between how these ad hoc CKEditor accordions look in the body (some whitespace seems unavoidable between accordions) vs. the how the embedded FAQ accordion was handled via Twig templating, where multiple accordions are flush next to each other. I think this is something that could be addressed later, just making a note of it.

Another thing to note is that the FAQ accordion button is wrapped by an H2 class="accordion-header" tag. I know we're making an effort toward semantic completeness and/or screenreader compatibility, so adding the H2 wrapper to this feature for consistency's sake may be worthwhile.

Last observation: accordions are being allowed in the FAQ content type since their Answer body is CKEditor enabled. Since embedded FAQ content will be commonplace, this could lead to accordions within accordions. Again, hopefully most of our editors have the sense not to abuse this, so, probably not a priority.

jmilden-git:
@lgbridges Do I need to do code review on this PR?

lgbridges:
> @lgbridges Do I need to do code review on this PR?

I've passed code review.

lgbridges (approved):

lgbridges (approved):

PR #302, opened 2021-12-16
WCOR-0: Restylize PDF embeds to allow listing.

jlwcor:
### ODE 234

**Use/Fix Case** 

As an author I would like to put PDF document embeds using the media entity library into lists.

**Test Plan**

- [x] Embed multiple PDF documents into a list. (After adding a media document to the first item in a list, use SHIFT+ENTER on your keyboard to break to the next line.)


sharstine:
@jlwcor @lgbridges - QA passes.

Page used for testing: /housing/qa-page

Here's what I tested:

* Embedded documents in list in one column
* Embedded documents in list in two columns
* A long document title (to ensure it wraps)
* A document description field exposed (to ensure it wraps)
* Desktop and mobile experience

It takes a bit of fiddling to insert a media document into a bullet point, especially if you insert the media document first and then convert it to a bullet point. If you start with a bullet point and then insert a media document, and then press Shift+Enter, the editor creates a line break (as expected) and doesn't create a new bullet point. To convert the second list item to a bullet point, I had to click the bullet point icon in the toolbar. I don't see a frequent use case for inserting media items into lists, so hopefully the training and support impact will be minimal.

When a list contains both media documents and regular text, the spacing between ```<li>```s without media documents is different than spacing between ```<li>```s that contain media documents.

lgbridges:
Made a couple small updates to style to accommodate the odd spacing in lists and mildly adjust icon placement.

lgbridges (approved):

PR #303, opened 2021-12-16
WCOR-0000: Add theme building and debugging to aid local development.

if-nick:
### ODE XX

**Use/Fix Case** 

Code updates resulting from stand up meeting. These add Lando and Drupal configuration helpful for site building. 

**Implementation Details**

- `.lando.yml` updated.
- This allows building the production version of the theme with a `lando build-theme`
- Or building the development version of the theme with a `lando watch-theme`
- Also, allows turning on Xdebug for local development and IDE debugging with a `lando xdebug-on`

**Dependencies**

- 

**Test Plan**

-


docroot/sites/default/settings.php
```diff
@@ -852,3 +852,24 @@
...
+ * Enable development configurations if in a local, developer environment.
+ *
+ * Here we can disable caching, show twig template suggestions,
+ * disable JavaScript/CSS compression, or re-route outgoing emails,
+ * etc, to help site building.
+ *
+ * Keep this block at the bottom of `settings.php` to override other variables.
+ * @see _developer_drupal_debug_configurations/settings.append.txt
+ */
+$file = __DIR__ . '/settings.local.php';
```
jlwcor, line 867: Line 798 already checks to see if settings.local.php exists and includes it if found. This should work for the lando environment as well. Does the variable $app_root come up as undefined in Lando?

jlwcor (changes requested):

xdebug.ini
```diff
@@ -0,0 +1,7 @@
+xdebug.mode = debug
+xdebug.client_host = ${LANDO_HOST_IP}
+xdebug.idekey = vscode-xdebug
```
joshuami, line 3: We may need to leave this out so that the Cloud IDE environments still have their correct xdebug settings. Maybe we could add this as a snippet to our wiki documentation. https://github.com/Washington-County/COUNTY-DRUPAL/wiki

joshuami (changes requested):

docroot/sites/default/settings.php
if-nick: Thank you for the catch. I haven't seen `app_root` come up undefined but will keep an eye out. The lines above have been removed.

xdebug.ini
if-nick: Thank you for the suggestion. I've added `xdebug.ini` to the git ignore file and removed it from this pull.

jlwcor:
@if-nick looks like c2f4c8b reverted back some of the changes that were made.
Ex. readding the local.settings.php in settings.php

jlwcor:
Replaced with #344

PR #314, opened 2021-12-21
WCOR-0: Bugfix to address schema-json vulnerability and deprecated calc.

joshuami:
### ODE 238

**Use/Fix Case** 

Dependabot has been reporting a schema-json vulnerability for some time. This is a bugfix to address that vulnerability, update theme dependencies and remove a deprecated approach to calculating a margin/padding value to use the calc() function instead.

**Implementation Details**

- `npm update`
- Refacator `$spacer / 2` to `calc($spacer / 2)`

**Dependencies**

- Not applicable

**Test Plan**

- [x] Quick smoke test of the theme on the CD environment. This is a low risk update.


joshuami:
@lgbridges and @jlwcor , I created you a quick hotfix to resolve the dependabot alerts that we keep getting regarding the json-schema vulnerability in packages.json. I should probably walk y'all through the theme maintenance steps that are similar to what we see with composer dependency updates. It wouldn't hurt for us to add a test in our pipelines that fails if there are any npm-audit results in the theme to help keep that all up to date. I'll add the the ODE value when the pipeline finishes running.

jlwcor (approved):

PR #334, opened 2021-12-23
WCOR-1226: Refine and style search results display.

lgbridges:
### Use case

As a user, I need a uniform, attractive display of search results including content type labels.

### Implementation details

Refine and style acquia_search view.

### Test Plan

- [x] Open /search and enter a search query. View the search results and see that each content type represented is uniform and contains the same search excerpt display. Confirm that each result has a content type label.
- [x] Open /news and enter a search query. View the search results for similar styling to /search.


sharstine:
@lgbridges - QA passes per test plan.

Note: The first several times I ran search queries on /search (using the terms "housing" and "landlord"), I was presented with an error and no results returned. I wasn't paying too much attention to the error itself because I was planning to go back and take a screenshot of it. Then I went to /news and ran the same two queries, and they both returned results with no error. So I went back to /search to run the queries again, and they worked. I tested many other queries and didn't get any errors after the initial failures, even when I tested in a different browser.

**Future enhancements to consider:**

- /search and /news - On mobile, pagination controls aren't wrapping. This creates empty space to the right of the search results listing and prevents content from extending the full width of the viewport. It's functional but could be better optimized for mobile visitors. MultCo and Portland both wrap their pagination controls on mobile: https://www.multco.us/search/site/application, https://www.portland.gov/search?keys=application&op=

- /search and /news - On mobile, could we display an expandable Filters link/area near the search field so visitors don't have to scroll to the bottom of the page to access the filters?

- /search - When filtering by one or more groups, such as Housing Services and Animal Services, it's not obvious which groups are currently selected because they're moved to the bottom of the Group filters list, out of view. If we have to move selected groups up/down the list, maybe we move them to the top rather than the bottom? (In general, I find it a little confusing that selections are moved at all. If I want to filter by two groups, I've already spotted them both in the list. After I select the first group, the second group has moved and I have to scan to find its new position. It's functional but creates a little more work for the visitor.)

- /news - Unable to filter by more than one year at the same time. If I select 2020, for example, all the other years disappear.

- /news - Unable to filter by more than one month at the same time. If I select 2020 and then October, for example, all the other months disappear.

lgbridges:
Ready for second round CR/QA.

@sharstine -- Made adjustments to address most of your great feedback. 

- Pagination controls should be right for mobile now.
- Short of adding filter controls near the search query field, I've at least set a jump link ("Set search filters", open to feedback on verbiage) that will take mobile users down to the facets (filters) at page bottom.
- Corrected the issue with facet sorts so that selected items stay on top.
- Added counts to facet options.
- News "Filter by year" facet will now accept multiple selections. However, have disabled "Filter by month" facet since I'm unsure if the built-in facet configuration can accommodate a conditional/hierarchical year/month interoperability. Given that our peer organizations don't seem to offer date filtering for news items, I'll want to discuss the utility of this with the group before investing further time.

sharstine:
@lgbridges - The adjustments look great.

Additional comments:

- /search and /news - If you want to filter by multiple years, content types, or groups, you can only make one selection at a time. For example, if you want to filter news by 2017-2021, you have to select 2017 and then click "Set search filters" again (on mobile) to be taken back to the filters to select 2018. And repeat for 2019, 2020, 2021. It's functional and I'm not sure how often visitors will want to select multiple years, content types, or groups at the same time. **This seems to be consistent with Portland.gov, so maybe we leave it as is.**

- /news - Unlike the filters on /search, the "Filter by year" values don't move to the top of the list when selected. For example, if you click Show More > 2010, the page reloads and there's no indication that 2010 is an active filter, especially since only the three most recent years are displayed under "Filter by year" before you click Show More. I think it's unlikely that visitors will search for content more than three years old, so **maybe we leave it as is for now and revisit later if there's a request to enhance it.**

- /news - Can only filter by one source. For example, if you select Sheriff's Office, the rest of the sources disappear and you can't select additional sources, unless the source you selected (e.g., HHS) is a parent of other sources (e.g., Public Health). In reality I can't see visitors wanting to filter by more than one source on a regular basis. As a workaround, they can deselect the currently filtered source and select a different source. **This is consistent with Portland.gov, so maybe we leave it as is.**

lgbridges:
Corrected the issues raised in 2nd round QA. Considering this complete.

jlwcor (approved):

PR #337, opened 2021-12-28
WCOR-1235: Enable card widgets in CKEditor.

jlwcor:
### ODE 243

**Use/Fix Case** 

As a content author, I would like to insert bootstrap cards into my content via CKEditor.

screenshot

**Implementation Details**

- Develop and install custom CKEditor card module.

**Dependencies**

- custom/ckeditor_accordion

**Test Plan**

- [x] Insert a card in the body using the Card tool button in CKEditor.


sharstine:

Tested as Eddie Editor: /hhs/qa-card-widgets

Editing the "Go somewhere" styled link text could be tricky for authors, and the experience is a little different in Firefox than in Chrome and Edge. We may want to consider making the styled link text (and possibly the href itself) editable via a modal, similar to how the card title is entered. This would hopefully eliminate the issues below.

- **In Chrome and Edge:** To edit the text of the styled link, you can simply click the styled text and begin typing to insert text, or select the portion of text you want to replace/remove using the mouse. Seems OK. 

- **In Firefox:** To edit the text of the styled link, you have to right-click the link, then left-click the link, then use the arrow keys on the keyboard to move the cursor. Using a mouse, you can't select/highlight or position the cursor to insert link text. Not intuitive for authors.

In all browsers:

- Clicking the Unlink icon in CKEditor removes the styling behind the link. We'll need to train authors to update, rather than remove, the link target.
- If you delete all characters in the link text except for the first one (the letter "G") and begin typing, the new text falls outside the ```<a>``` tag and isn't properly styled.

jlwcor:
@sharstine good points. Might have to remove the stylized link text as the default placeholder item.
There is a another PR coming that will assist with link styling.

PR #338, opened 2021-12-28
Activity/WCOR-19 Theme site navigation and header.

if-nick:
### [ode244](http://wcorode244.prod.acquia-sites.com/)

**Use/Fix Case** 

Style and theme site header. [GitLab Ticket](https://git.ifsight.net/proj/clients/wcor/wcor-issues-internal-01/-/issues/19) 

DRAFT pending review and feedback. 

**Implementation Details**

- Branched off of `spike/WCOR-0000`, lando debugging updates, and `activity/WCOR-22`, initial SCSS variables work to support the style tile. 
- Adds CSS source maps to `webpack.mix.js`, helpful for finding the source of style rules in aggregated CSS files.
- Separates header template blocks into a separate file, `masthead.html.twig`, to isolate those sections and make the  parent template easier to work with
- Avoids overriding radix contributed-theme templates.
- Extends minimal Bootstrap color utilities (although they are available); basic style guide colors can be used in templates with, for example `bg-faded-green`, one of our color variables as a CSS class.
- Uses plain JavaScript for header javascript rules, relying on Webpack and the theme build steps for polyfills.

NOTE:
- Does not include mobile/ break-points theming

QUESTIONS: 
- Should search form HTML be moved elsewhere, or to a new navigation template
- Should header.js use different selectors 
- Do we need additional ARIA work now, or review accessibility later
- Others

**Dependencies**

- [insert here]

**Test Plan**

- [ ] Header matches [Style Tile](https://projects.invisionapp.com/share/R511V1DRTHG7#/screens)
- [ ] Header matches [Style Guide](https://projects.invisionapp.com/share/NP123DYQ4UHM#/screens/461685912)
- [ ] Header should meet [WAVE Aria accessibility guidelines](https://wave.webaim.org/report#/https://wcorode244.prod.acquia-sites.com/cao/news/moratoria-de-desalojo-en-todo-el-estado) 
- [ ] Header logo should match design and be cacheable, optimized png (PENDING)
- [ ] Header viewed on different devices/ viewports should match designs and resize cleanly (PENDING)
- [ ] Header branding should match font sizes and colors 
- [ ] Main navigation should match font sizes, spacing and colors
- [ ] Main navigation drop-downs should match colors, spacing and have multiple columns if having 10+ items
- [ ] Main navigation search item should match colors, spacing, form borders and border radius of design


if-nick:
I intended to add labels for "draft" and "needs review" , and to tag some team members, but don't seem to have permissions to do so.

docroot/themes/custom/westy/westy.libraries.yml
```diff
@@ -5,6 +5,7 @@ style:
       assets/css/westy.style.css: {}
   js:
     assets/js/westy.script.js: {}
+    https://kit.fontawesome.com/fba24a4d16.js: { type: external, minified: true }
```
jlwcor, line 8: We are still exploring the option of using fontawesome.
In the meantime, are you able to use a css based approach for the icon? 
https://css.gg/search

jlwcor (changes requested):

joshuami:
@if-nick , was the form supposed to display on the CD environment for testing. I'm still seeing an oddity with the search form flyout.

image

joshuami (approved):

joshuami:
@if-nick , the login link should get moved to the footer as a config change:

image

Do we need to add a label to the search form? It feels like it should read fine, but it is throwing an alert in WAVE.

image

It's not caused by this story, but the content column acted a bit weird at tablet levels. @lgbridges , we should get a story to address that.

image

The test plan has items that have not been addressed yet. It would be better to move those into a follow up story if we aren't going to test all of those before merging.

image

jlwcor:
@if-nick can you provide a quick summary explanation on what is being changed in docroot/themes/custom/westy/webpack.mix.js ?

if-nick:
> @if-nick can you provide a quick summary explanation on what is being changed in docroot/themes/custom/westy/webpack.mix.js ?

There are two big changes in the webpack configuration- 

One checks if the current build is for production, as in Acquia environments or environments where a build is run with `mix --production`.  If not production we chained `sourceMaps` to the mix.  This enabled CSS source file mapping and would allow developers to inspect CSS to see what rules or files may be in use theming. 

The other was to add a SASS option for "quiet dependencies". The thinking there was that even though Bootstrap generates dozens of warning from the linter, they're not necessarily code and files we manage so the warning don't have lots of use. Reducing the amount of warnings should make identifying errors we do manage easier to spot and take care of.

if-nick:
> Do we need to add a label to the search form? It feels like it should read fine, but it is throwing an alert in WAVE.

I'm not sure if a <label>, hidden so it matches design, would be read by assistive technologies. I believe we're okay on this one and the WAVE documentation does say "If the form control is intuitive without a <label>, the title attribute value may be used."

if-nick:
> @if-nick , the login link should get moved to the footer as a config change

Thank you. Configuration updated and added

PR #353, opened 2022-01-08
WCOR-0: Fix differences in config export ordering.

jlwcor:
### ODE XX

**Use/Fix Case** 

There are config file differences in dev, stage, and prod due to the reordering of some config lines.

**Test Plan**

- [x] Verify no config differences exist in the ODE @ /admin/config/development/configuration
- [ ] Once merged, verify there are no config differences listed for all the environments at /admin/config/development/configuration


joshuami (commented):
@jlwcor , looks like this branch is out of date with develop. Otherwise the change looks correct.

PR #355, opened 2022-01-11
WCOR-1246: Add legacy path to Resource content type.

lgbridges:
### Use case

As a site admin, I need to be able to set legacy paths for Resource items, allowing legacy redirects against Resources that point users to things like (external) Forms Server pages or (internal) facet-filtered view-pages such as Group Documents.

### Implementation details 

- Enable the field for Resource form display. (Simpler than expected.)

### Dependencies

- wcor_legacy_redirects module

### Test plan

- [x] As site admin, create a test Resource item. Set a Legacy Path of your choosing (doesn't need to be a currently valid URL on www.co).
- [x] Append the legacy path to the ODE domain and submit in browser. If done as logged in user, like site admin, you will be directed to the Resource in its default display mode. (An anon user would simply be forwarded to the Resource URL.)


jlwcor (approved):
Approved. Just a note that /apps will not work since there is currently a directory in docroot that overrides redirects.

PR #359, opened 2022-01-12
WCOR-1248: Main header styling and block placement.

jlwcor:
### ODE 253

**Use/Fix Case** 

Stylize main header components and blocks per style guideline.

screenshot

**Implementation Details**

- Relocate title, breadcrumb, notification blocks to correct regions.
- Remove summary from general display view and replace with views block.
- Stylize main header dependent on node type.
- Add landing page title to end of breadcrumb navigation.
- Allow featured images to be displayed for group page headers.

**Test Plan**

- [x] Create a notification and confirm that it is on top of content body and is stylized.
- [x] Confirm new breadcrumb and main header styles are set for page, group, and news content types.
- [x] Upload an image to the featured image labeled field in groups. Image should be displayed in the header.


lgbridges:
@jlwcor -- Much of this is working great. There is some ground to cover concerning image embed in the header.

- I'm assuming that we would want the Group's Featured Image to be a media reference to an image in the media library instead of an image field attaching the image directly to the group node.
- On mobile, the image should stack above breadcrumb/title/summary, but I didn't see that yet.

image

jlwcor:
@lgbridges stacking is functional now on smaller view points.
As for upload method, we could use media references, but I'd like to keep files we use in templates vs files we use for content separate. Like how we use separate county logo files in the header than referencing it by media library. 

This ensures that deleting media library items does not affect the site's template.

joshuami:
I would recommend a media entity reference field for any featured image implementation.

Using a media entity instead of a file field gives you the ability to call the image into the page using a display mode—though technically you can still load a responsive image style with the image. A media entity references also allows reuse rather than uploading it for each one. Uploading per field means duplicating the file if it is reused—which has both storage and caching implications.

I'm not 100% sure how to dynamically serve up a responsive background image when the image is brought through as a field like this, but we should set aside some time (even if its phase2) to fully implement an image strategy that loads a smaller image on mobile devices. You page rank and your user's data plan will thank you for it 😉

jlwcor:
@joshuami @lgbridges the main issue is that all the images that get uploaded as a media entity gets dumped in the same directory with the rest of the site's images.

We shouldn't have someone using a photo sized for a featured-image banner as something to be used in the body field or vice versa. 

More importantly, for images being used in the templates, we need to set varying resolution and size restrictions to ensure ratio is standardized.

lgbridges:
> @joshuami @lgbridges the main issue is that all the images that get uploaded as a media entity gets dumped in the same directory with the rest of the site's images.
> 
> We shouldn't have someone using a photo sized for a featured-image banner as something to be used in the body field or vice versa.
> 
> More importantly, for images being used in the templates, we need to set varying resolution and size restrictions to ensure ratio is standardized.

I see both sides. There is some balance of extra work to be done under time pressure vs. flexibility down the line, and I'm not sure where it currently lands.

I'm under the impression that managing the images as media items means that we have more flexibility in presenting the images (as Josh mentions) using various display modes, in case anything needs to change in the future with the header layout, among other possible reasons. This would be the main reason to pursue this route.

Admins are the only ones that can update Group; therefore we're the only ones who will be linking media to Featured Image on Groups, thereby ensuring that ratio and resolution are correct.

Media items aren't deletable except by admins. So the question is: how do we ensure that **_we_** don't delete a template image? I'm assuming we could either manage this with views or EVA (some mechanism that tells us whether the image we're looking at is currently in a Group Featured Image field), or we could add Boolean as an admin-only field to images to indicate that it's not to be deleted.

We can keep these images out of the main Media Library, meaning only members of that group would see the images for potential re-use. We can't prevent re-use of the images, but it will obviously be clear to Publishers that the image is way wrong for body content.

I don't know if this helps.

lgbridges:
All that said, the test plan passes as it stands.

jlwcor:
> I'm under the impression that managing the images as media items means that we have more flexibility in presenting the images (as Josh mentions) using various display modes, in case anything needs to change in the future with the header layout, among other possible reasons. This would be the main reason to pursue this route.

The images styling is done through CSS. If the templates were to change down the line, we would do style through CSS. There is absolutely no reason to apply various image stylings to a static section of the site that doesn't not need various display modes. It is setup by default using TWIG and CSS to auto adjust the width and center the image as long as the correct resolution image is uploaded.

joshuami:
@jlwcor and @lgbridges , I can nearly guarantee that you will be impressed by the responsive image styles combined with display modes and focal point cropping. Having images sized by the system and loaded by breakpoint saves ton of bandwidth over styling images exclusively with CSS. The focal point crop also takes care of so much of what you are suggesting be done with photo editing software. It makes it possible to give over some additional control to your users as well. Here is an example of what the image preview looks like when it is loaded:

image

I hope to have some time next week to build out all of the image styles, responsive styles and basic display modes to cover the use cases in the style tiles. That should fix quite a few images that have already been embedded as well. You can choose whether or not to use it for featured images, but it would be a shame not to do so.

> Admins are the only ones that can update Group; therefore we're the only ones who will be linking media to Featured Image on Groups, thereby ensuring that ratio and resolution are correct.

Your group homepages have content and will need your publishers for the groups to be able to edit them to change that content. The won't necessarily need to edit the group to feature news, but likely you will want editing the group to be the path to featuring services and other static types of content.

> We can keep these images out of the main Media Library, meaning only members of that group would see the images for potential re-use. We can't prevent re-use of the images, but it will obviously be clear to Publishers that the image is way wrong for body content.

Because of the dynamic resizing the focal crop and display modes give, your editors could definitely reuse an image that was used in a heading for a program in a news item that they plan to promote in social media. Making the images reusable is what Drupal media is all about and it is very well supported between core an a couple of heavily-used-and-well-supported contrib modules.

jlwcor:
@joshuami again, all good stuff, but knowing our content authors, they won't use all these tools.
All they want is to upload an image and have it adjusted automatically for them to a fixed style we provide.

This has been the case for our current CMS and much of our satellite Drupal sites and it has worked just fine.

We're on a tight deadline to get the rest of the components for the site built out and I would prefer we don't spend more time on this. I'm open to revisiting this for phase 2, but like much of the current theme development, we are looking to make templates as static as possible and then open it up for more customizations little by little down the road.

PR #370, opened 2022-01-13
WCOR-1249: Alert component.

jlwcor:
### ODE 256

**Use/Fix Case** 

As a site admin, I would like to post alerts and have it displayed on the top of the site.

9ff692d5-d56d-45a9-9dd5-7679df81051b web ahdev cloud_

**Implementation Details**

- Configure alerts view.
- Create new alertSystem fuction.
- Develop new template and stylize.
- Relocate alerts to be on top of navbar.

**Test Plan**

- [x] Create an alert and ensure it is shown at the top of the page.
- [x] Close out the alert and refresh the page. The alert should no longer show for that particular browser session.


docroot/themes/custom/westy/src/js/westy.script.js
```diff
@@ -18,6 +18,37 @@ import './_header.js';
   //   }
   // };
 
+  // // Controls sitewide alert banner function.
```
joshuami, line 21: @jlwcor , one use case you might want to add in a follow up is bringing back alerts that have been updated so that the user sees changes and updates to those alerts. I've included some sample code below that would set a cookie that gets invalidated when the alert node changes. I'd also recommend grouping all of the alert code into a component in docroot/themes/custom/westy/src/components and calling that component into your view. If we keep adding to the westy.script.js, it could get a little unruly at some point. 

```
import $ from 'jquery';
import Drupal from 'Drupal';

const COOKIE_PREFIX = 'Drupal.visitor.cloudy_notification_dismissed.';

Drupal.behaviors.notificatin_handler = {
  /**
   * @param {HTMLElement} context
   * @param settings
   */
  attach(context, settings) {
    // Compare each server side alert changed time with browser cookie values.
    // If the changed time doesn't match for that alert, display the alert.
    $('.cloudy-notification').once('alert-processed').each(function() {
      // If this alert has no nid it is not dismissible and did not set a cookie
      if (!$(this).data('nid')) return;

      const nid = $(this).data('nid');
      const cookieChangedTimestamp = $.cookie(COOKIE_PREFIX + nid);
      const alertChangedTimestamp = $(this).data('changed');
      if (cookieChangedTimestamp != alertChangedTimestamp) {
        // Only show the alert if dismiss button has not been clicked. The
        // element is hidden by default in order to prevent it from momentarily
        // flickering onscreen.
        $(this).addClass('cloudy-notification--active-dismissible');
      }
    });

    // Set the cookie value when dismiss button is clicked.
    $('.cloudy-notification__close').click(function (event) {
      event.preventDefault();
      const alertElement = $(this).closest('.cloudy-notification');

      // Hide the alert
      alertElement.removeClass('cloudy-notification--active-dismissible');

      // Set an alert cookie
      const nid = alertElement.data('nid');
      const lastChangedTimestamp = alertElement.data('changed');
      const path = (drupalSettings && drupalSettings.path && drupalSettings.path.baseUrl) || '/';
      $.cookie(
        COOKIE_PREFIX + nid,
        lastChangedTimestamp,
        {
          path,
        }
      );
    });
  }
};
```

joshuami (commented):
Let me know if you want to tackle any of the comments in this PR or would prefer to tackle in a follow up.

config/default/views.view.alerts.yml
```diff
@@ -150,11 +150,76 @@ display:
           multi_type: separator
           separator: ', '
           field_api_classes: false
+        nid:
```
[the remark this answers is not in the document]
jlwcor: Removed

docroot/themes/custom/westy/src/js/westy.script.js
jlwcor: Added a epoch time of changed date to the uid so updated alerts reshow.

jlwcor:
@sharstine 

> * There are two character counters under the "Alert body" field. One of them updates in real-time and the other says 255 characters remaining all the time. A character limit doesn't seem to be enforced currently.
>   
>   * Content limited to 255 characters, remaining: -58
>   * Content limited to 255 characters, remaining: 255

The duplicate character count is a bug that arises when used on a body as it thinks a summary is also present. I've enabled force truncate so that it does not go beyond 255.

> * I was able to post 3 alerts concurrently. Should we limit the number of alerts that can be published at the same time, or will we just use our best judgment?

I'm going to leave this open for now as it's better to have the option of posting more than one available.

> * I'm not seeing any visual difference between Medium and High severity alerts. They both have the same icon and background color.

Waiting on IF design for differences as they only gave us one color.

PR #372, opened 2022-01-14
WCOR-1250: Create contact us component for group pages.

jlwcor:
### ODE 257

**Use/Fix Case** 

As a visitor, I would like to view contact information for groups.

9ff692d5-d56d-45a9-9dd5-7679df81051b web ahdev cloud_at (1)

**Implementation Details**

- Setup new view block for contacts. Use existing service location address EVA.

**Test Plan**

- [ ] Visiting group pages shows contact information on right sidebar.


docroot/themes/custom/westy/templates/views/views-view-unformatted--contact-info--group-contact.html.twig
```diff
@@ -0,0 +1,61 @@
...
+
+  {% endblock %}
+</div>
+
+<style>
+  .contact-item {
+    border: 0;
+  }
+  .contact-item .card-header {
+    background-color: #152E3A;
```
joshuami, line 49: There are bootstrap classes for styling card text and headings. (https://getbootstrap.com/docs/5.0/components/card/) We should be judicious in adding custom classes and overrides as that adds CSS that will contribute to total download size.

docroot/themes/custom/westy/templates/views/views-view-unformatted--contact-info--group-contact.html.twig
```diff
@@ -0,0 +1,61 @@
...
+    <div class="card-header">
+      <h2>Contact Us</h2>
+    </div>
+    <div class="card-body">
+      {% for row in rows %}
+        {% set phone = row.content['#view'].style_plugin.render_tokens[ loop.index0 ]['{{ field_phone }}']|render|striptags|trim %}
+        {% set phone_ext = row.content['#view'].style_plugin.render_tokens[ loop.index0 ]['{{ field_phone_extension }}']|render|striptags|trim %}
+        {% if phone %}
+          <p class="card-text">
+            <strong>Phone</strong><br>
```
joshuami, line 15: It would be better to use the field output with field labels where we can. Hardcoded text like this doesn't get the `t()` function for translation and might differ from the field label used elsewhere.

docroot/themes/custom/westy/templates/views/views-view-unformatted--contact-info--group-contact.html.twig
```diff
@@ -0,0 +1,61 @@
...
+            {{ phoneString(phone, phone_ext) }}
+          </p>
+        {% endif %}
+
+        {% set serviceuid = row.content['#view'].style_plugin.render_tokens[ loop.index0 ]['{{ field_primary_service_location }}']|render|striptags|trim %}
+        {% set address = drupal_view('service_location_display', 'address_contact_card', serviceuid)|render|striptags('<br><p>')|trim %}
+        {% if address %}
+        <p class="card-text">
+          <strong>Address</strong><br>
```
joshuami, line 24: Since a service location is not really a mailing address for a group, we discussed displaying this as either a location card or adding a mailing address field to the contact information. While this matches the design, the design doesn't match the content model's intention to provide a get directions link when service location information is presented.

joshuami (commented):
One last thought on the approach of theming views with twig rather than theming entity types and display modes. I'm pretty sure calling all these fields instead of an entity loses a layer of caching. The render array wants to save the results for anything it can. If all we call are fields, only the entire view is cached rather than each result in the view. I'm still trying to wrap my head around the performance implications since I've seldom used twig templates for views—other than maybe to move the view header or to place the table caption region.

docroot/themes/custom/westy/templates/views/views-view-unformatted--contact-info--group-contact.html.twig
jmilden-git: This isn't really a card component.  This is a call out box with title bar, I think custom classes are appropriate.  But I agree for the promotional cards using the bootstrap classes is best.

docroot/themes/custom/westy/templates/views/views-view-unformatted--contact-info--group-contact.html.twig
jmilden-git: Yes, I have a similar issue with that.  We are missing additional service location information.

jmilden-git:
QA results:
I see the Contact box style on a few pages, but I'm not sure what node this is applied to.  There is only one "contact" type content in the entire system.  There are also other Service Locations on pages that don't have the Contact box style applied.  
@jlwcor can you explain where this will appear?

docroot/themes/custom/westy/templates/views/views-view-unformatted--contact-info--group-contact.html.twig
```diff
@@ -0,0 +1,61 @@
...
+        </p>
+        {% endif %}
+      {% endfor %}
+    </div>
+  </div>
+
+  {% endblock %}
+</div>
+
+<style>
```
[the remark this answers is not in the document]
jlwcor: Using `|render` still caches fields as evident in our contacts list story: https://github.com/Washington-County/COUNTY-DRUPAL/pull/251#issuecomment-962269785 

The issue with display modes in this case is the twig template is unable to retrieve field values that are disabled. There are many cases where we need the value of a field but don't necessarily want it displayed.

jlwcor:
> QA results: I see the Contact box style on a few pages, but I'm not sure what node this is applied to. There is only one "contact" type content in the entire system. There are also other Service Locations on pages that don't have the Contact box style applied. @jlwcor can you explain where this will appear?

@jmilden-git this only applies for groups as there is a direct contact field.
Contact cards for other content types are coming in a later PR.

docroot/themes/custom/westy/templates/views/views-view-unformatted--contact-info--group-contact.html.twig
jlwcor: I'm going to keep this in for now and remove it when we get the service location card components built out. That way we can address duplicate addresses being shown on page (ie. contact card and service location card.)

docroot/themes/custom/westy/templates/views/views-view-unformatted--contact-info--group-contact.html.twig
jlwcor: Agreed.

PR #373, opened 2022-01-14
WCOR-1251: Enable Group Accordions CKEditor widget.

jlwcor:
### ODE 258

**Use/Fix Case** 

As a content author I would like to group accordions in addition to creating single accordions.

9ff692d5-d56d-45a9-9dd5-7679df81051b web ahdev cloud_cao_news_moratoria-de-desalojo-en-todo-el-estado
screenshot

**Implementation Details**

- Create new ckeditor_groupaccordion module.

**Test Plan**

- [ ] Insert a group accordion via CKEditor. By default it should generate two accordions.
- [ ] Insert additional accordions into the group via the single accordion tool.


sharstine:

### General observations
To insert additional accordions into the group, I had to click the red "Insert paragraph here" icon under one of the existing accordions in the group. Then "Insert Accordion" from the CKEditor toolbar, and then remove the paragraph break directly above the newly inserted accordion. I could see this process being confusing for authors, although I don't think authors will be inserting accordions very often.

The only difference I see between the group accordion and inserting accordions individually is the bottom margin on accordions that are inserted outside a group. Since the group accordion only inserts two accordions by default, I wonder if it's simpler to have just the single accordion tool (i.e., no group accordion) and authors can still add as many individua accordions as they need. If the reason for a group accordion was discussed elsewhere, please disregard as I probably missed it.

### Bug (maybe)
Here's an issue I encountered in **_Chrome_**. Ultimately not a blocker after I discovered I was trying to insert additional accordions the wrong way:

1. Click at the very top or bottom of the group accordion region.

bug

2. Click "Insert Accordion" in the CKEditor toolbar.
3. Enter a title.
4. Click OK.

Nothing happens. An error is recorded in the Chrome DevTools console.

```
Uncaught TypeError: Cannot read properties of undefined (reading 'checkReadOnly')
    at $.insertElementIntoRange (ckeditor.js?v=4.17.1:366:45)
    at $.insertElement (ckeditor.js?v=4.17.1:364:502)
    at $.<anonymous> (ckeditor.js?v=4.17.1:375:34)
    at b.d (ckeditor.js?v=4.17.1:10:190)
    at b.<anonymous> (ckeditor.js?v=4.17.1:12:32)
    at b.window.CKEDITOR.window.CKEDITOR.dom.CKEDITOR.editor.CKEDITOR.editor.fire (ckeditor.js?v=4.17.1:13:225)
    at b.insertElement (ckeditor.js?v=4.17.1:276:436)
    at CKEDITOR.dialog.onOk (accordion.js?t=r5opq6:90:16)
    at CKEDITOR.dialog.<anonymous> (ckeditor.js?v=4.17.1:609:97)
    at CKEDITOR.dialog.d (ckeditor.js?v=4.17.1:10:190)
```
In **_Firefox_**, the modal goes away when you click OK and no errors are recorded in DevTools, but the accordion is never added to the group.

### Minor enhancement suggestion
Add bottom margin to group accordion.

group-accordion-no-bottom-margin

jlwcor:
> Nothing happens. An error is recorded in the Chrome DevTools console.

I was able to replicate your error. Not necessarily a bug, but it's due to clicking on space that doesn't actually exist. You'll need to use the red "insert paragraph" tool to create a space below and insert an accordion in that empty paragraph.

PR #382, opened 2022-01-18
WCOR-1253: Primary Service Location Card

jlwcor:
### ODE 260

**Use/Fix Case** 

Design and create primary service location card for groups.

9ff692d5-d56d-45a9-9dd5-7679df81051b web ahdev cloud_hhs (1)

**Implementation Details**

- Create custom view block and retrieve service location info and reference building info.

**Test Plan**

- [x] Group pages (Ex. /hhs) properly display their primary service location info in the sidebar.


sharstine:

Observations and potential issues:

- Street address shows up in the right sidebar twice -- under "Contact Us" and in the service location card. Should we get rid of the street address under "Contact Us"?
- I added an image to service location /charles-d-cameron-public-services-building/. The image I uploaded doesn't show up in the service location card on /hhs. I'm still seeing the placeholder image.
- When the service location has no image, can we hide the photo region in the service location card?
- The hours could be misleading on County-observed holidays. I like how Google puts a message on business profiles near the hours on holidays: "XXXX holiday might affect these hours." Maybe when a County-observed holiday is approaching, we display a "Check holiday schedule" link that points to the service location page. This feels like a phase 2 item to me, though. Not an immediate need.

jlwcor:
> The main purpose of Location in Groups, then, would just be to provide the physical address for the Contact Us card. 

@lgbridges makes sense. I'll go ahead and omit if from groups.

jmilden-git:
@jlwcor @lgbridges Please don't remove this yet until we've had a chance to discuss this in a meeting.  The content has a need for location in places.  The only thing I'd point out is the redundancy of having the street address in both the Contact box and the Location box.  It makes more sense to keep the street address in a Location card.

lgbridges:
> @jlwcor @lgbridges Please don't remove this yet until we've had a chance to discuss this in a meeting. The content has a need for location in places. The only thing I'd point out is the redundancy of having the street address in both the Contact box and the Location box. It makes more sense to keep the street address in a Location card.

@jmilden-git -- Having two different kinds of contact/location cards appearing on group homepage is redundant, that's what we're eliminating here. The service location card and the contact card are separate use cases. Service Location specifically applies to service pages and news.

I think street address is still a useful thing to have on the Contact card that will appear on Group homepage, but the Contact card makes it more abstract for the purpose of presentation on a group homepage.

jmilden-git:
@lgbridges Someone has modified the Group pages in Production and removed the location information.  This was part of the content.  Let's not remove content from pages in production.  If you want to mock up an idea for the CD env, that seems like a safer approach.

jlwcor:
> Someone has modified the Group pages in Production and removed the location information
@jmilden-git @lgbridges I specifically brought up in our meeting yesterday that the primary_field_location would get removed and with it the data. The data has been placed back into the new location_field hence the recent modification. There is no missing location data at the moment... It's just not being shown on the page display.

PR #386, opened 2022-01-19
WCOR-1246: Update system date formats.

lgbridges:
### Use/Fix Case

Date formats need to be normalized and made more consistent with web content from our peer organizations. 

(This will not have a lot of current impact without datetime display on most content right now. It will largely be noticed as a change in admin views.)

### Implementation Details

- System default date formats updated. 
- Content admin view updated.

image

### Test Plan

- Review the date formats at /admin/config/regional/date-time.
- Check system view /admin/content and note that both date & time are displayed for Updated value. (This is a new format called Default short date & time).
- Check system view /admin/content/media and note that only short date is displayed for Updated value. (This is the original short format, but the timestamp has been removed.)


composer.lock
```diff
@@ -18911,5 +18908,5 @@
     "prefer-lowest": false,
     "platform": [],
     "platform-dev": [],
-    "plugin-api-version": "2.1.0"
+    "plugin-api-version": "2.2.0"
```
jlwcor, line 18911: Set this back to 2.1.0

jlwcor (changes requested):

jlwcor:
@lgbridges everything checks out in CR.

jlwcor (approved):

PR #394, opened 2022-01-20
WCOR-1243: Content review / approval workflow (with scheduling).

lgbridges:
### Use case

As an editor, I need to be able to submit content for publisher review.

As a publisher, I need flexible, intuitive admin views to approve content in review, and audit recent approvals for my group. I need scheduling capability to work in conjunction with the review process.

As an admin, content moderation workflows, scheduling, and revisions functionality require adjustment for compatible operation and production readiness.

### Implementation details

- Permissions adjustments
  - Remove editor ability to revert revisions
  - Enable view revisions permissions for editor through content admin
  - Align workflow state transitions with expected role access
- Refinement of moderation workflows
  - Remove all redundant transition aspects (same origin and destination state).
  - Editorial workflow changes
    - Rename _Create new draft_ transition to _Return for changes_.
    - Remove Editor permission to _Return for changes_; create new _Save draft_ transition just for Editors.
    - Add Purge draft transition exclusively for Editor use. (Publishers have _Mark for deletion_.)
    - Remove all types except contact, FAQ, news, notification, and page from Editorial workflow.
    - Add _Scheduled_ workflow state and _Schedule publication_ transition to accommodate scheduling.
  - Media workflow remains unattached to any media type (discuss implications with team).
- _My content_ view pivots to _**My group content**_, since it shows all content in the logged in user's group (including those not created/updated by the user). (Same for _My media_ -> _My group media_.)
- _**My group content**_ also serves as the moderation queue, as it allows Publishers to:
  - View "In review" items for potential approval
  - Switch to viewing items by other workflow states Scheduled, Published, Unpublished, and Marked for deletion (in the case of Scheduled or Published, this allows them to see who approved the latest revision of an item).
- **_My content_** will then be a brand new view specific to the Editor/Publisher to show items they created (Authored by), who updated them, and their publication / review status.
- Several adjustments to streamline content admin views.
  - Correct Manage Content links in _My groups_.
  - Remove unused fields from _My groups_.
  - Updates to _My content_ and _My group content_.
- Install Diff module for comparing revisions (from Revisions tab of node).
- Test scheduling functionality with workflows.

### Dependencies

- Modules
  - Content moderation
  - Diff
  - Workflows

### Test plan

- [x] Masquerade as Eddie Editor. Add a new page and populate fields with test values. _Displayed In_ value should be "Information Technology Services". For the final _Save as_ field, choose "In review". Save the page. Unmasquerade.
- [x] Masquerade as Polly Publisher. Visit the _My group content_ view-page (/my-group-content). Find the page just created by Eddie and click **Edit** on the right end of the row.
- [x] Find the _Scheduling options_ block on the right sidebar and expand it. Set _Publish on_ date and time to a future time (a few minutes from now). **Be sure** to set _Publish State_ in this block to Published. At the bottom of the page, set the workflow state change field _Change to_ to Scheduled. Unmasquerade.
- [x] As admin, run cron after the time designated for scheduled publication. (See screenshot.)
image
- [x] View the page as public user (anon) to ensure all changes are committed to the published revision.
- [x] Masquerade as Eddie Editor. Visit the _My content_ view-page (/my-content). Find the page that was just created (and published by Polly) and click **Edit** on the right end of the row. Make a change to the Body field. Save the page under "In review" status. Unmasquerade.
- [x] Masquerade as Polly Publisher. Visit the _My group content_ view-page (/my-group-content). Find the page Eddie just edited and note the Moderation state of "In review (has published revision)". Click **Edit** on the right end of the row.
- [x] Before taking action in the Edit tab, click the Revisions tab. Compare the latest revision in review (select radio button in left column) with the currently published revision (select radio button in right column). Review the differences. Press the browser Back button to return to the Edit tab. Change workflow state to Published and save. Unmasquerade.
- [x] View the page as public user (anon) to ensure all changes are committed to the published revision.
 
### Notes

Media currently does not have revisioning, which presumably means a media entity can't be updated with a new media file to reduce the amount of work and duplication needed to update media. We need to consider adding this to the backlog.


sharstine:
@lgbridges - Thank you for the well-written and easy-to-follow test plan. QA passes as test plan is written.

Questions and suggestions:

1. I think we should require Editors to fill out the "Revision log message" field. A brief summary of the change will help Publishers in their review and will provide descriptive history on the Revisions tab. However, I don't think "Revision log message" should be required for Publishers who are reviewing and changing the state of a page. Also, I see Editors easily overlooking "Revision log message" because it's nowhere near the "Change to" field and Save button.

2. If an Editor goes to edit an existing page and clicks "Save" without actually making any changes, the state of the page changes to "In review" regardless. There's no "Cancel" button. I suppose we can train Editors to click the Back button in the browser if they decide to not save/submit changes. Or we could make the "Revision log message" field required to prevent non-changes from being submitted.

3. There's currently no way for an Editor to save changes to an existing page without submitting the page for review. There may be cases when an Editor needs to make significant updates to a page in multiple sit-down sessions. Or maybe the Editor needs to step away for a while and wants to save his/her work, but isn't ready for a Publisher to review the changes. A "Draft" state for revisions to existing content could be helpful.

4. Will there be a way for a Publisher to refer back changes to an Editor? Would the Publisher just change the state to "Draft" and leave a note in the "Revision log message" field? Would the Editor receive any sort of notification outside of Drupal about the page being referred back to them? Or would the Publisher be responsible for modifying any content issues they find and educating the Editor outside of Drupal?

5. Related to the above, are there any email notifications that will go out when a page is put into "In review" status?

6. Under "Scheduling options," would it make sense to put the "Publish" fields above the "Unpublish" fields? A page would go live before it's later unpublished, so in my mind the fields should be in that order.

7. On the Revisions tab, the two radio buttons on each row are confusing to me. They're not labeled so I don't know the difference between clicking one or the other.

8. Do we really want to allow pages to be scheduled for unpublishing? I'm having a hard time thinking of a valid use case. I would think unpublishing should be a manual action. Better yet, maybe we could redirect an unpublished page to a designated replacement page or the nearest parent page. Or offer a more graceful message on the unpublished page than "Access denied." Perhaps a new story idea would be to define a process and/or find a solution for expiring content in Drupal.

9. The "Resource" content type currently has no approval workflow. I think that makes sense. Just pointing it out in case it was an oversight.

10. I'm able to create a "Contact" as an administrator, but I get a "Page not found" when creating a new "Contact" as Eddie Editor. Could be unrelated to this PR.

lgbridges:
> @lgbridges - Thank you for the well-written and easy-to-follow test plan. QA passes as test plan is written.
> 
> Questions and suggestions:
> 

Thanks for the awesome feedback, @sharstine.

Combined with feedback from today's sprint retro demo, there are a few immediate tweaks I can make before we merge this. Anything that isn't a quick fix will probably go into another upcoming story. Will probably open a Discussion on a couple of these items that may need broader team feedback. Again, as always, appreciate your eye for detail in the QA work.

lgbridges:
I've addressed QA bulleted issues 3, 6, and 8 with the latest commits. Most of the rest will kick over to a [refinement story in the backlog](https://github.com/Washington-County/COUNTY-DRUPAL/discussions/403).

lgbridges:
Added default value selection for Publish State in Scheduling Options.

jlwcor (approved):

PR #397, opened 2022-01-21
WCOR-19: Correct style and behavior of mobile main menu

if-cpacheco:
### ODE 266

**Use/Fix Case** 

- Corrects style and behavior of mobile main menu

**Implementation Details**

- Went with a simpler mobile approach and behavior than mocks which utilizes the least amount of custom templating

**Dependencies**

- N/A

**Test Plan**

- [ ] Test breakpoints for desktop and mobile


docroot/themes/custom/westy/templates/page/page.header.html.twig
```diff
@@ -1,5 +1,3 @@
-<div class="container">
```
jlwcor, line 1: @if-nick we will probably want to put this back into the container so that it doesn't extend edge to edge on a super wide screen.

docroot/themes/custom/westy/src/scss/layout/_page.header.scss
```diff
@@ -33,6 +42,20 @@ header.page__header {
...
+    }
+
+    padding: 1rem 0;
+
+    me-auto {
+      padding: 1.5rem 0 !important;
+    }
+    h1 {
+      @include media-breakpoint-down(lg) {
+        @include visually-hidden;
```
jlwcor, line 56: Based on renderings, should this be moved to the bottom as opposed to hidden?
https://projects.invisionapp.com/share/NP123DYQ4UHM#/screens/461685922

jlwcor (changes requested):
Looking better. A few changes requested.

PR #401, opened 2022-01-21
WCOR-1265: "How Do I" homepage component.

jlwcor:
###  [ODE 269](http://wcorode269.prod.acquia-sites.com/)

**Use/Fix Case** 

Create baseline "How Do I" homepage component.

screenshot

**Implementation Details**

- Hardcode component into front page twig.
- Custom component CSS.

**Dependencies**

- Icons to come.
- Header font-weights to come.

**Test Plan**

- [x] "How Do I" component is visible on homepage.


docroot/themes/custom/westy/src/scss/components/_howdoi.scss
```diff
@@ -0,0 +1,82 @@
...
+  .nav-tabs {
+    background-color: rgb(236 247 254 / 85%);
+    padding: 0 1.25rem;
+
+    h3 {
+      margin: 0;
+    }
+
+    .nav-item .nav-link {
+      background-color: inherit;
```
lgbridges, line 17: An easy step toward consistency with Style Guide would be to set background-color to none for .nav-link buttons (all states, see also line 25). There are other possible suggestions, but it sounds like a follow-up story is pending.

lgbridges: If you do that, button.active h3 should be also set color:#000 (can't find where atm).

jlwcor: @lgbridges applying it to all styles changes the node edit buttons for logged in users.
For now I've made it global as this is the better implementation.

PR #405, opened 2022-01-24
WCOR-1253-A: Enable featured photo for service location cards.

jlwcor:
### ODE 271

**Use/Fix Case** 

Follow up on #382 to implement featured images in service location card components. 

> I added an image to service location /charles-d-cameron-public-services-building/. The image I uploaded doesn't show up in the service location card on /hhs. I'm still seeing the placeholder image.

> When the service location has no image, can we hide the photo region in the service location card?

**Implementation Details**

- Use existing featured image field to populate image in view.

**Test Plan**

- [x] In an existing page, select a location for it and save. Visiting the page should render a service location card on the right sidebar.
- [x] Edit the service location content and select a featured image and save. Visiting the page edited in the first task should now render a service location card with a featured image.


sharstine:

Related observations -- I believe these are both by design to prevent duplicate information in the sidebar:

- The service location card renders on pages (when the Location field is populated), but not on groups.
- The "Contact Us" block renders on groups (and includes an address when the Location field is populated), but not on pages.

PR #408, opened 2022-01-24
WCOR-1259: WCSO homepage/theme build starter.

jlwcor:
### [ODE 272](http://wcorode272.prod.acquia-sites.com/AH_VIEW)

**Use/Fix Case** 

Initiate theme override for WCSO.

**Implementation Details**

- Utilize theme preprocess to check and verify if URL is /sheriff/* and output true/false statement for twig use using `{{ wcor }}`

**Test Plan**

- [ ] Visiting /sheriff displays the WCSO landing page.
- [ ] Visiting /sheriff/about-us displays the page with certain CSS overrides (Ex. top nav bar bottom border colors is yellow.)


docroot/themes/custom/westy/templates/block/block--system-branding-block.html.twig
```diff
@@ -13,19 +13,33 @@
...
+            {{ 'Sheriff'| t }}
+            <span>{{ 'Washington County'| t }}</span>
+          </a>
+        {% else %}
+            {{ 'Washington County'| t }}
+            <span>{{ 'Oregon'| t }}</span>
+          </a>
+        {% endif %}
+
       </h1>
```
lgbridges, line 43: It looks like the Style Guide is calling for placement of [their branding image](https://ecomm.co.washington.or.us/Crywolf_Online/local/images/logo_topleft.png) instead of text. Current PR could work as a draft, but at some point we may need to update to include the image instead of text. (The H1 could then be hidden for WCSO for semantic completeness.)

jmilden-git:
@jlwcor  QA is failing since the site is not displaying.  Please let me know when you are ready for QA.

docroot/themes/custom/westy/templates/block/block--system-branding-block.html.twig
jlwcor: @lgbridges replaced it with the WCSO text logo.
Thanks for that. Minor adjustments with colors and padding here and there, but this PR should be a good start to get all that merged in for future PRs.

PR #414, opened 2022-01-26
WCOR-1256: Promo content type.

jlwcor:
### [ODE 274](https://wcorode274.prod.acquia-sites.com/)

**Use/Fix Case** 

Enable promos to be displayed on county and sheriff homepages.

**Implementation Details**

- Create new promo content type.
- Setup views and custom block to retrieve assigned promos.

**Test Plan**

- [x] Visit '/' and verify that the 'News & Updates' block readjusts with no promo content.
- [x] Visit '/sheriff' and verify that the 'News & Updates' block readjusts with no promo content.
- [x] Create a new promo content and set placement to county homepage. Verify it shows in '/'
- [x] Using the same created promo, now set the placement to sheriff homepage and save. Verify it shows in '/sheriff'


lgbridges:
@jlwcor Will share some feedback momentarily. I have some concerns about sustainability for this that I'd like to review with you.

lgbridges:
Some suggestions here -- less concern after fully reviewing but some things to mention.

Firstly, my assumption is that the new Promo content type concept is primarily here for:

- Selecting a featured image for the main promo spot
- Selecting placement on either County or WCSO homepage

If Page type had a field_featured_image field, we could simply pull the image from the related content, meaning we wouldn't need this unique field. _But_ to my knowledge we can't easily, in a view block, extrapolate the parent/grandparent group for field_displayed_in. That would mean that currently, we do need the field_placement in order to enable the Promo, so it seems like there is a role for the Promo type.

One thing that needs to be fixed:

- **"Learn more" link href** contains an "entity:" prefix that needs to be removed for the link to work. Also, the actual alias/path name would be preferable to display here, if possible.

I would also suggest the following changes to streamline this implementation:

- Remove Body field from the Promo form. Body content won't appear in the promo space. Summary would be displayed, ideally, and that should be pulled from the underlying content into the view block.
- Ideally, Promos view would display the latest published content revision date from the underlying content. This is especially relevant if additional updates happen to the referenced content during the period of Promo publication.

config/default/views.view.promos.yml
```diff
@@ -0,0 +1,496 @@
...
+          field: field_resource_url
+          relationship: none
+          group_type: group
+          admin_label: ''
+          plugin_id: field
+          label: ''
+          exclude: false
+          alter:
+            alter_text: true
+            text: '{{ field_resource_url__uri }}'
```
lgbridges, line 232: Something besides URI may be need to avoid the "entity:" link prefix that prevents clickthrough.

lgbridges (changes requested):

config/default/views.view.promos.yml
jlwcor: Working on a fix for this.

jlwcor:
> Remove Body field from the Promo form. Body content won't appear in the promo space. Summary would be displayed, ideally, and that should be pulled from the underlying content into the view block.

@lgbridges remember that we are keeping promos open to link to anything at this point. Let's say for example there needs to be a promo that links to some survey outside the county site, we would still need to utilize the body field. 

Ideally we wouldn't want promos to depend on anything from another content node in the case it's external.

lgbridges:
> > Remove Body field from the Promo form. Body content won't appear in the promo space. Summary would be displayed, ideally, and that should be pulled from the underlying content into the view block.
> 
> @lgbridges remember that we are keeping promos open to link to anything at this point. Let's say for example there needs to be a promo that links to some survey outside the county site, we would still need to utilize the body field.
> 
> Ideally we wouldn't want promos to depend on anything from another content node in the case it's external.

@jlwcor If an external URL is deemed important enough to be a basis for front page Promo, it needs to be created as a Resource first and given a summary. Only something of summary length should be allowed in the Promo block.

lgbridges:
I would also strongly recommend adding a Sort to the Promos view to sort on Promo updated date. Since only one Promo is allowed per homepage (County/WCSO), if more than one Promo was left published by the responsible role, a sort by date will be needed to clarify which one should be published (the most recently updated one).

jlwcor:
> Only something of summary length (and without markup) should be allowed in the Promo block.

@lgbridges this can be applied in the view. It is already removing HTML tags. 

Again, for Phase 1, I still think the best approach is to keep things open and slowly implement restrictions here and there as we being to better understand how the content workflow develops.

Like the workflow for FAQs, it makes it tedious for authors to create two different content types (resource and promo) have both be approved at the same time, for it to display properly in a placement location. Resources were primarily developed to reference a single item multiple times in content. This does not apply to homepage promos.

jlwcor:
@lgbridges latest commit adds 3 fixes:

- Now sorted by last changed date.
- Truncates body, appending '...' to 255 characters max.
- URL now fixed to use redirect path.

lgbridges:
> > Only something of summary length (and without markup) should be allowed in the Promo block.
> 
> @lgbridges this can be applied in the view. It is already removing HTML tags.
> 
> Again, for Phase 1, I still think the best approach is to keep things open and slowly implement restrictions here and there as we begin to better understand how the content workflow develops.
> 
> Like the workflow for FAQs, it makes it tedious for authors to create two different content types (resource and promo) have both be approved at the same time, for it to display properly in a placement location. Resources were primarily developed to reference a single item multiple times in content. This does not apply to homepage promos.

I agree with keeping things simple, but I also want a sufficiently dynamic content management experience for our County stakeholders.

But if we are making Promos something that Editors can propose but that a higher role has to approve, then you are right about the double-hop issue. I was assuming that Editor-level authors might not be involved in Promo management.

If you are going to localize the text content to the Promo type, is there any reason not to at least swap body for field_summary so there's a basic length restriction in place? You can strip HTML tags from body, but authors may not be expecting that and could result in confusion. Just wondering why we wouldn't use field_summary when we're not looking to put much text in that space anyway.

jlwcor:
> I was assuming that Editor-level authors might not be involved in Promo management.
@lgbridges Eventually it would be good to see promos make it's way into other pages, specifically group landing pages, but the current design comps don't accommodate for that scenario so it's something we'll have to implement for phase 2. If that does get built, we'll need to see who can make those promos. 

> If you are going to localize the text content to the Promo type, is there any reason not to at least swap body for field_summary so there's a basic length restriction in place?

I'll go ahead and make the swap.

lgbridges (approved):

PR #417, opened 2022-01-27
WCOR-1266: Sidebar nav updates.

lgbridges:
### Use case

As a site visitor:

- I need a streamlined local/sidebar nav experience that doesn't include too many options, especially options that aren't relevant to my specific information need, such as siblings of the current page that are largely unrelated except for sharing the parent page or group that sponsors them.
- I need an additional "Return to parent" breadcrumb link at the bottom of the page to give additional visibility and power to breadcrumbs as a means of navigating content.

As a site admin, I need Content region for pages and groups to span a majority width of the page. I also need to ensure that navigation menus contain at least two child items to appear.

### Implementation Details

This story shifts the orientation of sidebar nav to displaying only child menu items (instead of showing both siblings and children). It also places the nav on the right sidebar, effectively leaving the left sidebar without operational blocks and therefore collapsed for every displayed node. Presenting sidebar blocks on the right for desktop browsers is in keeping with the desired presentation order for mobile browsers.

#### Why eliminate siblings from menus and only show children?

- It greatly reduces the amount of cognitive demand needed by users on a page, decreases the amount of overwhelm as they review the page and its nav. Reducing the number of unhelpful navigation options is in keeping with accessibility best practices.
- The content model should drive toward longer pages of closely related content. In cases where a model of more "atomic" pages is preferred by editors, there needs to be more robust linking of body content to connect these shorter pages to weave sibling page content together organically. This needs to be discussed in editor training. 
- Related field for pages, as well as group homepage, also need to be used strategically to present related content instead of relying on a comprehensive sibling-based nav.

Additionally, sidebar nav with only one child item will not be displayed. Site editors will need to add at least two child menu items to a page in order for nav to appear. The need for page content consolidation in a single child item scenario needs to be addressed in editor training.

### Test Plan

As site visitor (anon):
- [x] Visit a page or group at division level or lower (e.g. /mcfp) and verify that the sidebar nav is in the right sidebar below Contact Us. Confirm that only child pages appear.
- [x] Visit any page and confirm that a new "Return to [parent]" link appears at the bottom of body content.

As site admin:
- [x] Create a new test page. In Menu Settings, check _Provide a menu link_ and choose a page or group as a parent. 
_(Hopefully this works in ODE, there were issues in local with select list data population, but it appeared to be an IDE issue independent of this PR work.)_
- [x] Repeat the prior step, giving the page a different title and short title. Choose the same menu parent as you did before.
- [x] Visit the page or group chosen as the menu parent for the test pages and confirm that your test pages appear in sidebar nav. Ensure that only child pages appear in sidebar nav.


composer.lock
```diff
@@ -19226,5 +19226,5 @@
     "prefer-lowest": false,
     "platform": [],
     "platform-dev": [],
-    "plugin-api-version": "2.1.0"
+    "plugin-api-version": "2.2.0"
```
jlwcor, line 19229: giphy

Annoying composer!

jlwcor (changes requested):

composer.lock
lgbridges: Dang. I will fix.

sharstine:
@lgbridges - QA passes as test plan is written, with one exception: Eddie Editor doesn't have permissions to the menu settings on pages. I tested the Eddie Editor functions as an administrator instead.

Comments:

1. /housing/aloha-inn-permanent-supportive-housing: Suggest adding spacing between the Translation block and the menu in all viewports.

translation_menu_spacing

2. In the "< Return to [parent]" area at the bottom of pages, the spacing between "to" and "[parent]" is slightly more than the spacing between "Return" and "to." Can we reduce the spacing before [parent]?

return-to-parent-spacing

3. It appears that the "Contact Us" block is only displayed on groups. While unrelated to this PR, I think there are cases when it would be helpful to display group contact information on pages as well: https://github.com/Washington-County/COUNTY-DRUPAL/discussions/369.

jlwcor (approved):

PR #422, opened 2022-02-02
WCOR-1271: Group 'Contact us' block for pages.

lgbridges:
### Use case

As an editor, I need to make sure contact information for a page's associated group(s) is displayed on the page.

### Implementation details

Implement additional view-block, implement toggle field for Page to display the block, restyle block. Confirm compatibility with multiple **_Displayed in_** values for page.

### Test plan

As Eddie Editor:

- [x] Edit any page. Toggle on the _Display contact information block_ checkbox (directly below _Displayed in_). Save and review the page to ensure the contact block appears in the sidebar as expected.
- [x] Edit the page again. Populate two blocks into _Displayed in_. Save and ensure contact block appears as expected.
- [x] Ensure other cards on site appear as expected, with consistent styling, padding, etc (example: promo card on County homepage).


lgbridges:
@sharstine -- ODE is up & ready for QA, thanks.

jlwcor (commented):
This approach seems a bit confusing as we already have a 'contacts' field for pages.
A cleaner approach would be to include groups as selectable entity references for the contacts field and combine both contact type info and group type info into one contact us block.

That would remove the need to have an additional "show block" field and allow folks to also display individual, non-group level contacts in the same area.

jlwcor (changes requested):
This also breaks the style of existing cards.

screenshot

sharstine:
@lgbridges - QA passes.

The only minor suggestion I have is to add some spacing between "Return to [parent]" / "more news" and the Contact Us block on mobile viewports.

return_to

more_news

lgbridges:
> This approach seems a bit confusing as we already have a 'contacts' field for pages. A cleaner approach would be to include groups as selectable entity references for the contacts field and combine both contact type info and group type info into one contact us block.
> 
> That would remove the need to have an additional "show block" field and allow folks to also display individual, non-group level contacts in the same area.

@jlwcor -- I understand what you're going for but they seem like different contexts. We've got the following separate use cases:

- Individual contact (a person, generally) in Contacts field. Outputs in Contacts field below body content in Content region.
- Multiple contacts (people) in Contacts field. Converts the page (via twig logic) to displaying a list of contacts as content.
- (NEW) Group contact derived from Displayed In, toggled on by new checkbox. Displays in Page sidebar in same way as on every Group.

This latest one was an attempt to match what we see on Contact Us on Group pages -- same card style (but allowing for multiple groups if multiple Displayed In values, including secondary header for that).

We're kinda assuming users will be keeping an eye on the top sidebar location for a contact block since that's already happening in Group. I prefer that current location for this card -- I wouldn't want to output group contact at the bottom of the page.

Also, it may be a bit more expensive on the backend to introduce a display toggle field, but if a page's Contacts field were to accept Group references, then it could presumably accept any Group. This feels like an odd pattern somehow (to me). The current implementation ensures that only Displayed In groups are included in Contact Us card.

lgbridges:
> This also breaks the style of existing cards.
> 
> screenshot

I will rework styles to address this.

jlwcor:
@lgbridges contacts and groups share similar contact info fields.
If someone wants to display an important contact information like a project manager for a page, we should display it in the same contact us card. We would just need to allow contacts to be references in the contact field as well.

If someone has multiple contacts (>2) or something, then they should use the contact list display.

This would be easy to implement and fit into your design. See example: 


screenshot

lgbridges:
> @lgbridges contacts and groups share similar contact info fields. If someone wants to display an important contact information like a project manager for a page, we should display it in the same contact us card. We would just need to allow contacts to be references in the contact field as well.
> 
> If someone has multiple contacts (>2) or something, then they should use the contact list display.
> 
> This would be easy to implement and fit into your design. See example:
> 

@jlwcor -- I definitely like how this consolidates the existing page-level contacts display into the card format. However, it does expand the scope of this story.

Contact type is going to have to expand and change. To fit the example you shared where the card's secondary header isn't the name of an individual but their _role_ (which I think makes sense), we would have to make _Contact title_ field (maybe that should say "position title") required to populate that. 

We might additionally consider making the _Name_ of the individual optional. Mitzi's work is showing us that there are unofficial groups and committees within the County (no official Group in Drupal) that only have a group name to associate with the Contact and not just one individual.

Also, if these changes are made and accepted, I imagine there may need to be some changes to the Twig for the Contact-type pages that display multiple contacts in a table view, etc.

lgbridges:
@jlwcor -- I agree that the enhancements we discussed would be beneficial, but I am going to move on to other higher priority story work for now. I have noted some observations in #436 for us to further discuss, scope, and refine this for a follow-up story.

lgbridges:
@sharstine -- Please see the additional QA item in test plan and check when you have a moment.

sharstine:
@lgbridges - QA passes. Other cards on the site (i.e., promo card on the County/WCSO homepage and cards constructed via the "Insert Card" tool in CKEditor) appear to have the proper styling in different browsers and viewports.

PR #427, opened 2022-02-02
WCOR-1245: Embed documents inline instead of via block.

lgbridges:
### Use case

As an editor, I need to be able to create ordered/unordered lists of documents without extra spacing.

### Implementation details

- Update twig.

### Test plan

As Eddie Editor:

- [ ] Create a test page. Add an unordered list with stub text to create at least two items. Return to each of those items and replace the stub text with an embedded document. Save the page.
- [ ] Review the page and ensure no additional spacing appears between list items.


jlwcor:
@lgbridges wondering what the need for this is as @jmilden-git is already able to do listed documents:

https://wcordev.prod.acquia-sites.com/finance/2021-22-county-budget-reports

lgbridges:
> @lgbridges wondering what the need for this is as @jmilden-git is already able to do listed documents:
> 
> https://wcordev.prod.acquia-sites.com/finance/2021-22-county-budget-reports

You're right--I'm not seeing a difference at this point in how the embed appears in a <ul> right out of the gate, whether it's in a <div> or <span>.

I can close this unless anyone can provide a use case for inline doc embeds.

lgbridges:
Closing since no use case or added utility has been identified.

PR #435, opened 2022-02-03
WCOR-1266: split group versus page menu approach and refacor to show active trail.

joshuami:
### ODE XX

**Use/Fix Case** 

Show only children on group entities and only siblings on pages. Show active trail and highlight the current page for pages.

**Implementation Details**

- New group-specific menu block and new logic in menu template.

**Dependencies**

- NA

**Test Plan**

- [ ] Click through a few departments, divisions, programs, and their pages. There should be few examples of 10 or more menu items and mostly relevant child or sibling listings. (Some additional follow up on menu structure will still be needed, but it should be minor.)


jlwcor:
Down the line or even on this PR, we might want to revert the sidebar menu styling back to the way it was in the original renderings. Current blue style was developed for differentiating between multi-level nav items.

PR #438, opened 2022-02-07
WCOR-1287: Featured service block for groups.

jlwcor:
### [ODE 288](http://wcorode288.prod.acquia-sites.com/)

**Use/Fix Case** 

Allow for services to be featured for group pages.

screenshot

**Implementation Details**

- New services field for groups to reference entities.
- Feature services referenced with custom template.

**Test Plan**

- [ ] Visit an existing group and reference services. They should populate on the group's page.


lgbridges:
@jlwcor -- Basic functionality seems good. I'm assuming there will be a followup story (maybe this sprint?) to handle the view that provides all services via the "View services" link.

The main suggestion I have is to add a filter that can dynamically, using contextual filter/relationship, filter down the list of available service pages to those displayed in the group being edited. This can be a bit tricky in Views, but it would result in less margin for error when Groups are being edited to add featured services.

I can pass this for QA in the interest of time, but I definitely see this as a valid followup story.

jlwcor:
> The main suggestion I have is to add a filter that can dynamically, using contextual filter/relationship, filter down the list of available service pages to those displayed in the group being edited. This can be a bit tricky in Views, but it would result in less margin for error when Groups are being edited to add featured services.

@lgbridges I gave this some thought and keeping it open seems like a better way forwards. There might be a need for two departments to feature a single service page.

